PR #17, opened 2024-07-30
Added an example for Connector SDK

fivetran-satvikpatil:
Closes https://fivetran.height.app/T-742626

### Description
Added an example for connector SDK which uses pandas Dataframes to process and upsert the data.

### Testing
- `fivetran debug` output
Screenshot 2024-07-30 at 1 22 39 PM
Screenshot 2024-07-30 at 1 22 26 PM

- `Connector creation` output (Pending)
Screenshot 2024-08-04 at 9 27 12 PM
Screenshot 2024-08-04 at 9 27 21 PM

### DIFF
--- a/examples/user_profiles/connector.py
+++ b/examples/user_profiles/connector.py
@@ -6,7 +6,8 @@
 # Import required classes from fivetran_connector_sdk
 from fivetran_connector_sdk import Connector
 from fivetran_connector_sdk import Operations as op
-import requests as rq # Import the requests module for making HTTP requests, aliased as rq.
+# Import the requests module for making HTTP requests, aliased as rq.
+import requests as rq
 import pandas as pd
 
 # Define the schema function which lets you configure the schema your connector delivers.
@@ -14,6 +15,8 @@ import pandas as pd
 # https://fivetran.com/docs/connectors/connector-sdk/technical-reference#schema
 # The schema function takes one parameter:
 # - configuration: a dictionary that holds the configuration settings for the connector.
+
+
 def schema(configuration: dict):
     return [
         {
@@ -40,6 +43,8 @@ def schema(configuration: dict):
 # - configuration: dictionary containing any secrets or payloads you configure when deploying the connector.
 # - state: a dictionary containing the state checkpointed during the prior sync.
 #   The state dictionary is empty for the first sync or for any full re-sync.
+
+
 def update(configuration: dict, state: dict):
     # Retrieve the last processed profile ID from state or set it to 0 if not present
     profile_cursor = state["profile_cursor"] if "profile_cursor" in state else 0
@@ -47,22 +52,21 @@ def update(configuration: dict, state: dict):
     # Fetch new data and return DataFrames for profile, location, and login tables
     profile_df, location_df, login_df, cursor = get_data(profile_cursor)
 
-
     # Approaches to upsert the records from dataFrame
 
-    # APPROACH - 1
+    # APPROACH 1: Gives you direct access to individual row values by column name, Slower approach, helpful for custom row handling
     # UPSERT all profile table data, checkpoint periodically to save progress. In this example every 5 records.
     for index, row in profile_df.iterrows():
         yield op.upsert("profile", {col: row[col] for col in profile_df.columns})
         if index % 5 == 0:
             state["profile_cursor"] = row["id"]
             yield op.checkpoint(state)
-    
+
     # Checkpointing at the end of the "profile" table data processing
     state["profile_cursor"] = cursor
     yield op.checkpoint(state)
 
-    # APPROACH - 2
+    # APPROACH 2: Generally faster and more memory-efficient, Simplifies the code since rows are already dictionaries and can be used directly
     # UPSERT all location table data.
     # Iterate over each row in the DataFrame, converting it to a dictionary
     for row in location_df.to_dict("records"):
@@ -72,12 +76,12 @@ def update(configuration: dict, state: dict):
     state["location_cursor"] = cursor
     yield op.checkpoint(state)
 
-    # APPROACH - 3
+    # APPROACH 3: Faster approach, Keeps track of the original indices of the rows, which can be useful for certain operations that require indexing information.
     # UPSERT all login table data.
     # Iterate over the values of the dictionary (which are the DataFrame rows)
     for value in login_df.to_dict("index").values():
         yield op.upsert("login", value)
-    
+
     # Checkpointing at the end of the "login" table data processing
     state["login_cursor"] = cursor
     yield op.checkpoint(state)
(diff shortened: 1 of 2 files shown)


examples/user_profiles/connector.py
```diff
@@ -0,0 +1,144 @@
...
+from fivetran_connector_sdk import Operations as op
+from fivetran_connector_sdk import Logging as log
+import requests as rq
+import pandas as pd
+
+
+def schema (configuration: dict):
+    return [
+        {
+            "table": "profile",
```
alsnkline, line 23: Do we need to declare all the columns and data types?

This is an area customers are thinking is a requirement and is causing their code to take longer to write.
Customer are also not realizing we support a JSON data type.

I'm wondering if we could put a comment instead of declaring all the collumns in one of these tables - saying "Column and data type to be inferred by Fivetran"

Also could we make one of the columns in one of the declared tables of type JSON - so we cover that functionality also?

alsnkline (commented):
couple of ideas to be considered - not must haves

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,144 @@
...
+        }
+    ]
+
+
+def update(configuration: dict, state: dict):
+    profile_cursor = state["profile_cursor"] if "profile_cursor" in state else 0
+
+    profile_df, location_df, login_df, cursor = get_data(profile_cursor)
+
+    # UPSERT all profile table data and checkpointing after every 5 records
```
varundhall: comments are not consistent, please fix them.

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,144 @@
...
+                "profileId": "INT",
+                "street": "STRING",
+                "city": "STRING",
+                "state": "STRING",
+                "country": "STRING",
+                "postcode": "STRING"
+            }
+        },
+        {
+            "table": "Login",
```
fivetran-niketkhandelwal: this should be `login`

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,144 @@
+from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Operations as op
+from fivetran_connector_sdk import Logging as log
+import requests as rq
+import pandas as pd
```
fivetran-niketkhandelwal, line 11: add a `requirements.txt` for these modules

fivetran-niketkhandelwal (changes requested):

examples/user_profiles/connector.py
ediril: > Do we need to declare all the columns and data types?
No, we don't. Whatever columns the customer does not explicitly declare a type for will be inferred. As long as they are happy with the inferred type, no need to explicitly declare a type.

> so we cover that functionality also?
We should list all available types in the docs somewhere. They are in `common.proto` but connector sdk hides that from the customer

examples/user_profiles/connector.py
ediril: This isn't necessarily invalid. How would we handle this in a Java connector? We would apply renaming rules right?

examples/user_profiles/connector.py
ediril: No need to add comment if the code is clear. If the code is not clear, let's make it clear

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,143 @@
...
+
+def update(configuration: dict, state: dict):
+    profile_cursor = state["profile_cursor"] if "profile_cursor" in state else 0
+
+    profile_df, location_df, login_df, cursor = get_data(profile_cursor)
+
+    # UPSERT all profile table data and checkpointing after every 5 records
+    for index, row in profile_df.iterrows():
+        yield op.upsert("profile", {col: row[col] for col in profile_df.columns})
+        if index % 5 == 0:
```
ediril, line 61: This seems a bit contrived, why are you checkpointing like this?

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,143 @@
...
+            state["profile_cursor"] = row["id"]
+            yield op.checkpoint(state)
+    # Checkpointing at last
+    state["profile_cursor"] = cursor
+    yield op.checkpoint(state)
+
+    # Upserting data in "location" table in below loop
+    for index, row in location_df.iterrows():
+        yield op.upsert("location", {col: row[col] for col in location_df.columns})
```
ediril: This is not a bad example, but would it not be better to have a version of upsert that acceps a dataframe so the customer doesn't have to write the looping code over and over?

ediril (commented):
I added some comments

examples/user_profiles/connector.py
alsnkline: Listing all supported data types in the docs is the second item in [RD-740391](https://fivetran.atlassian.net/browse/RD-740391)
"In Technical Reference add information about every data type we support - people are not finding that we support JSON from the current datatypes example."
@varundhall has picked that ticket up.

examples/user_profiles/connector.py
alsnkline: Having this as a minimum example is good while we decide if we are going to add a helper method or similar to accept a pd.DataFrame directly.

examples/user_profiles/connector.py
alsnkline: Doesn't core apply the renaming rules?
I don't think we should cover that edge case in this examples - I would support renaming this in the example to 'login' to avoid the question here.

examples/user_profiles/connector.py
alsnkline: We should develop best practices for checkpointing - document them and then follow them in all our examples.

examples/user_profiles/connector.py
alsnkline: Users have explicitly requested more verbose comments in our examples.
ChatGPT comments every line and is setting expectations for this type of thing now.

ediril (approved):

examples/user_profiles/connector.py
fivetran-satvikpatil: I wrote this logic just to specify that they should perform a checkpoint after reading some number of records. Even though I can just checkpoint it at last, we shouldn't do that.

alsnkline: We could tweak the comment:
```
# UPSERT all profile table data, checkpoint periodically to save progress. In this example every 5 records.
```

examples/user_profiles/connector.py
fivetran-satvikpatil: Added verbose comments to example.

examples/user_profiles/connector.py
fivetran-satvikpatil: Added a column with JSON datatype.

examples/user_profiles/connector.py
fivetran-satvikpatil: I have added three different approaches to upsert the records from DataFrame. Please take a look.

varundhall (approved):

examples/user_profiles/connector.py
```diff
@@ -0,0 +1,150 @@
...
+    # APPROACH - 3
+    # UPSERT all login table data.
+    # Iterate over the values of the dictionary (which are the DataFrame rows)
+    for value in login_df.to_dict("index").values():
+        yield op.upsert("login", value)
+    
+    # Checkpointing at the end of the "login" table data processing
+    state["login_cursor"] = cursor
+    yield op.checkpoint(state)
+
```
alsnkline, line 88: Should we give some indication of when to use which approach?
are their any pros/cons for the different approaches - is one faster than the others?

alsnkline (approved):
Love it lets get this one out there.

examples/user_profiles/connector.py
fivetran-satvikpatil: @alsnkline Added some information about the approaches in the comments, please let me know if it looks good.

fivetran-niketkhandelwal (approved):

PR #18, opened 2024-08-02
Add verbose comments to examples

fivetran-niketkhandelwal:
Height Ticket: [Add ‘Chat GPT’ level verbose comments to all examples](https://fivetran.height.app/T-735408)
This PR adds comments to explain each line of code in detail


examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
+# It shows the use of a requirements.txt file and a configuration.json file to pass a credential key into the connector code and use it to decrypt a message
+# to get the weather forecast data for Myrtle Beach in South Carolina, USA.
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details.
+
+import json  # Import the json module to handle JSON data.
+
+# Import the Fernet class from the cryptography module for encryption and decryption. The cryptography module is listed as a requirement in requirements.txt
+from cryptography.fernet import Fernet
 from fivetran_connector_sdk import Connector
```
alsnkline, line 11: Should we add a comment
```suggestion
# Import required classes from fivetran_connector_sdk
```

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
 from fivetran_connector_sdk import Operations as op
-from cryptography.fernet import Fernet
```
alsnkline, line 3: We should import logs also
and add a FINE log - as we expect everyone to want to use logging in every connector

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
 
     return [
-            {
-                "table": "crypto",
-                "primary_key": ["msg"],
-            }
+        {
+            "table": "crypto",  # Name of the table in the destination.
+            "primary_key": ["msg"],  # Primary key column(s) for the table.
+        }
```
alsnkline, line 34:
```suggestion
          # table columns will be inferred based on the data sent
                }
```

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
     ]
 
 
+# Define the update function, which is a required function, and will be used to perform operations in the connector.
```
alsnkline:
```suggestion
# Define the update function, which is a required function, and is called by Fivetran during each sync.
```

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
 
+    # Yield an upsert operation to insert/update the decrypted message in the "crypto" table.
     yield op.upsert(table="crypto", data={
-        'msg': message.decode()
+        'msg': message.decode()  # Decode the decrypted message to a string.
     })
 
 
+# Instantiate a Connector object from the Connector class, passing the update and schema functions as parameters.
+# This creates a new connector that will use these functions to define its behavior.
```
alsnkline:
```suggestion
# This creates the connector object that will use the functions defined in this connector.py file.
```

is that technically correct? - I don't want to say a "new connector" because that will be confused by these users with the connection in the dashboard that is only created when they run the `fivetran deploy --....` command.

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
 connector = Connector(update=update, schema=schema)
 
-
 if __name__ == "__main__":
+    # Open the configuration.json file and load its contents into a dictionary.
```
alsnkline, line 70: can we add a comment here explaining what __main__ is and when it is used?
something like
```
# Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button. this is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production. Please test using the Fivetran debug command prior to finalizing and deploying your connector.
```

examples/configuration/connector.py
```diff
@@ -1,40 +1,67 @@
...
     with open("configuration.json", 'r') as f:
```
alsnkline, line 71: Is this actually needed?
I thought connector.debug(configuration=configuration) was sufficient?

examples/hello/connector.py
```diff
@@ -1,15 +1,32 @@
...
+# The state dictionary is empty for the first sync or for any full re-sync
 def update(configuration: dict, state: dict):
-    yield op.upsert("hello",
-                    {"message": "hello, world!"})
+    # The yield statement returns a generator object.
+    # This generator will yield an upsert operation to the Fivetran connector.
+    # The op.upsert method is called with two arguments:
+    # - The first argument is the name of the table to upsert the data into, in this case, "hello".
+    # - The second argument is a dictionary containing the data to be upserted,
+    yield op.upsert("hello", {"message": "hello, world!"})
```
alsnkline: I think this example to use the following format
```
yield op.upsert(table="hello", data={"message": "hello, world!"})
```
so its clear what is it doing

examples/hello/connector.py
```diff
@@ -1,15 +1,32 @@
...
 
 
+# Instantiate a Connector object from the Connector class, passing the update function as a parameter.
+# This creates a new connector that will use the update function to run updates.
```
alsnkline: see comment on prior example - we need to avoid using "a new connector" as that will be confused with the object create when you run fivetran deploy .....

alsnkline (changes requested):
I've made a number of comments - we should propagate them through the other examples - then I will review again.

thank you so much.

examples/configuration/connector.py
fivetran-niketkhandelwal: yes it's required. we need to load the json from the file into a python dictionary object to pass into the connector.debug command

examples/configuration/connector.py
```diff
@@ -1,40 +1,75 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# It shows the use of a requirements.txt file and a configuration.json file to pass a credential key into the connector code and use it to decrypt a message
+# to get the weather forecast data for Myrtle Beach in South Carolina, USA.
```
alsnkline: This line isn't correct for the configuration example

examples/hello/connector.py
```diff
@@ -1,15 +1,38 @@
...
+    # The yield statement returns a generator object.
+    # This generator will yield an upsert operation to the Fivetran connector.
+    # The op.upsert method is called with two arguments:
+    # - The first argument is the name of the table to upsert the data into, in this case, "hello".
+    # - The second argument is a dictionary containing the data to be upserted,
+    log.fine(f"upserting to table 'hello'")
+    yield op.upsert(table="hello", data={"message": "hello, world!"})
 
 
+# This creates the connector object that will use the update and schema functions defined in this connector.py file.
```
alsnkline:
```suggestion
# This creates the connector object that will use the update function defined in this connector.py file.
# This example does not use the schema() function, if it did it would need to be included in the connector object definition. 
```

examples/hello/connector.py
```diff
@@ -1,15 +1,38 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# It defines a simple `update` method, which upserts some data to a table named "hello".
```
alsnkline, line 2:
```suggestion
# It defines a simple `update` method, which upserts some data to a table named "hello".
# This example is the simplest possible as it doesn't define a schema() function, however it does not therefore provide a good template for writing a real connector.
```

examples/local/connector.py
```diff
@@ -1,44 +1,77 @@
...
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
+
+# Import required classes from fivetran_connector_sdk
 from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Logging as log
 from fivetran_connector_sdk import Operations as op
 
-
+# Define the SOURCE_DATA which simulates the source data that will be upstarted to Fivetran.
```
alsnkline:
```suggestion
# Define the SOURCE_DATA which simulates the source data that will be upserted to Fivetran.
```
typo in upserted

examples/local/connector.py
```diff
@@ -1,44 +1,77 @@
...
-                "primary_key": ["id"],
-                "columns": {
-                    "message": "STRING",
-                },
-            }
+        {
+            "table": "hello_world",  # Name of the table in the destination, required.
+            "primary_key": ["id"],  # Primary key column(s) for the table, optional.
+            "columns": {  # Definition of columns and their types, optional.
+                "message": "STRING",  # Contains a dictionary of column names and data types
```
alsnkline, line 31: Should we call out that the data type are optional?
should we call out that any columns omitted will be inferred?

examples/local/connector.py
```diff
@@ -1,44 +1,77 @@
...
 
+# Define the update function, which is a required function, and is called by Fivetran during each sync.
+# See the technical reference documentation for more details on the update function
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update
+# The function takes two parameters:
+# - configuration: dictionary contains any secrets or payloads you configure when deploying the connector
+# - state: a dictionary contains whatever state you have chosen to checkpoint during the prior sync
+# The state dictionary is empty for the first sync or for any full re-sync
 def update(configuration: dict, state: dict):
+    # Retrieve the cursor from the state to determine the current position in the SOURCE_DATA.
```
alsnkline:
```suggestion
    # Retrieve the cursor from the state object to determine the current position in the SOURCE_DATA.
```

examples/local/connector.py
```diff
@@ -1,44 +1,77 @@
...
-
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
 if __name__ == "__main__":
+    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE:
     result = connector.debug()
     if result:
         print("Success! You can publish your code now if you like: "
```
alsnkline: do we want this code in our examples:

if result:
        print("Success! You can publish your code now if you like: "
        
If so it should be in all of them - I'm not sure it adds much I'm tending to thinking we should remove it.

examples/unspecified_types/connector.py
```diff
@@ -1,39 +1,70 @@
...
+# The schema function takes one parameter:
+# - configuration: a dictionary that holds the configuration settings for the connector.
 def schema(configuration: dict):
     return [
         {
-            "table": "unspecified",
-            "primary_key": ["id"],
+            "table": "unspecified",  # Name of the table in the destination.
+            "primary_key": ["id"],  # Primary key column(s) for the table.
+            # No columns are defined, meaning the types will be inferred.
```
alsnkline, line 22: Column name will be taken directly from the data sent and the data type will be inferred - right??

I was left wondering where the column name would come from.

examples/unspecified_types/connector.py
```diff
@@ -1,39 +1,70 @@
...
-                        "_xml": "<tag>This is XML</tag>",
-                        "_str": "This is a string",
-                        "_json": {"a": 10},
-                        "_null": None
-                    })
+    # Yield an upsert operation to insert/update the row in the "unspecified" table.
+    # The data dictionary contains various data types.
+    # Since the schema does not specify column types, they will be inferred from the data.
+    log.fine("upserting to table 'unspecified'")
+    yield op.upsert(
```
alsnkline, line 40: be clear where the column names will come from

examples/unspecified_types/connector.py
```diff
@@ -1,39 +1,70 @@
...
-
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
 if __name__ == "__main__":
-    result = connector.debug()
\ No newline at end of file
+    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE.
+    result = connector.debug()
```
alsnkline, line 71: Should be include some ascii version of what table will result from this example?

examples/three_operations/connector.py
```diff
@@ -1,34 +1,63 @@
...
 connector = Connector(update=update, schema=schema)
 
-
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
 if __name__ == "__main__":
+    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE:
     connector.debug()
```
alsnkline, line 63: Should be include some ascii version of what table will result from this example?

examples/specified_types/connector.py
```diff
@@ -1,58 +1,82 @@
...
-
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
 if __name__ == "__main__":
-    result = connector.debug()
\ No newline at end of file
+    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE.
+    result = connector.debug()
```
alsnkline, line 82: Should be include some ascii version of what table will result from this example?

examples/weather/connector.py
```diff
@@ -1,69 +1,98 @@
...
 
     for period in periods:
-        # Skip data points we already synced
+        # Skip data points we already synced by comparing their start time with the cursor.
         if str2dt(period['startTime']) < str2dt(cursor):
             continue
-        
-        # This log message will only show while debugging
+
+        # This log message will only show while debugging.
```
alsnkline, line 68: We should call out that this examples demonstrates our logging functionality and/or put the FINE and INFO logs items into some of our other examples also.

Its very valuable functionality.

varundhall (commented):
@fivetran-niketkhandelwal can you please link the ticket and add some PR description.

examples/configuration/connector.py
fivetran-niketkhandelwal: fixed

examples/unspecified_types/connector.py
fivetran-niketkhandelwal: added a comment to explain that

examples/unspecified_types/connector.py
fivetran-niketkhandelwal: added a comment

examples/local/connector.py
fivetran-niketkhandelwal: right, removed

examples/local/connector.py
fivetran-niketkhandelwal: added a comment saying that any columns omitted from the schema will be inferred

fivetran-niketkhandelwal: @alsnkline - wdym by `data type are optional`? we must provide a dictionary of `column_name : valid_fivetran_data_type`, else it'll break

examples/unspecified_types/connector.py
fivetran-niketkhandelwal: done

examples/three_operations/connector.py
fivetran-niketkhandelwal: done

examples/specified_types/connector.py
fivetran-niketkhandelwal: done

examples/weather/connector.py
fivetran-niketkhandelwal: added a code comment. Do you think we should also document the same here: https://fivetran.com/docs/connectors/connector-sdk/detailed-guide#examples?

varundhall (approved):

examples/local/connector.py
alsnkline: Oh interesting.
I thought we could provide the column name without the data type. If I'm mistaken then this is fine.
Thank you for the confirmation

examples/weather/connector.py
alsnkline: 100% its bullet 3 in [RD-740391](https://fivetran.atlassian.net/browse/RD-740391) which @varundhall has been working on and is now in review.

alsnkline:
This looks great lets ship it :-) after a pass by tech writing.

examples/weather/connector.py
varundhall: @alsnkline yes, please have a look at that PR.

examples/configuration/connector.py
```diff
@@ -1,40 +1,82 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# It shows the use of a requirements.txt file and a configuration.json file to pass a credential key into the connector code and use it to decrypt a message
```
5tran-alexil:
```suggestion
# It shows the use of a requirements.txt file and a configuration.json file to pass a credential key into the connector code and use it to decrypt a message.
```

examples/hello/connector.py
```diff
@@ -1,15 +1,48 @@
...
+    # This generator will yield an upsert operation to the Fivetran connector.
+    # The op.upsert method is called with two arguments:
+    # - The first argument is the name of the table to upsert the data into, in this case, "hello".
+    # - The second argument is a dictionary containing the data to be upserted,
+    log.fine(f"upserting to table 'hello'")
+    yield op.upsert(table="hello", data={"message": "hello, world!"})
 
 
+# This creates the connector object that will use the update function defined in this connector.py file.
+# This example does not use the schema() function, if it did it would need to be included in the connector object definition.
```
5tran-alexil:
```suggestion
# This example does not use the schema() function. If it did, it would need to be included in the connector object definition. 
```

examples/local/connector.py
```diff
@@ -1,45 +1,83 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# It uses a local SOURCE_DATA object created at the start of the script, and demonstrates the Schema() method use.
+# It also shows a way to manage state.
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
```
5tran-alexil:
```suggestion
# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details.
```

5tran-alexil (approved):
LGTM

PR #25, opened 2024-08-20
docs(sdk): updated with supported python versions

manjutapali:
- Updating the supported python version in the docs


varundhall (approved):
please link a ticket and add some description
Also we should update the fivetran docs

PR #28, opened 2024-09-09
refactor(Example): Handling of exception from the update method

fivetran-satvikpatil:


examples/user_profiles/connector.py
```diff
@@ -46,45 +47,53 @@ def schema(configuration: dict):
...
+
+        # Checkpointing at the end of the "login" table data processing
+        state["login_cursor"] = cursor
+        yield op.checkpoint(state)
+
+    except Exception as e:
+        exception_message = str(e)
+        stack_trace = traceback.format_exc()
+
+        detailed_message = f"Error Message: {exception_message}\nStack Trace:\n{stack_trace}"
```
fivetran-satvikpatil, line 95: Added a try-catch block and included the stack trace along with an exception message when an exception is caught.

fivetran-satvikpatil:
Not needed

PR #29, opened 2024-09-11
feature(connector_sdk): Key based replication example

fivetran-rishabhghosh:
Example which shows Key based replication from DB sources


examples/key_based_replication/connector.py
```diff
@@ -0,0 +1,154 @@
...
+    return [
+        {
+            "table": "customers",  # Name of the table in the destination.
+            "primary_key": ["customer_id"],  # Primary key column(s) for the table.
+            "columns": {  # Define the columns and their data types.
+                "customer_id": "INT",  # Integer column for the customer_id.
+                "first_name": "STRING",  # String column for the first name.
+                "last_name": "STRING",  # String column for the last name.
+                "email": "STRING",  # String column for the email.
+                "updated_at": "UTC_DATETIME",  # UTC date-time column for the updated_at.
```
manjutapali, line 38: mention this column is considered as replication key in the comment

examples/key_based_replication/connector.py
```diff
@@ -0,0 +1,154 @@
...
+from datetime import datetime
+# import duckdb to interact with DuckDB databases from within your Python code.
+import duckdb
+
+# Import required classes from fivetran_connector_sdk
+from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Logging as log
+from fivetran_connector_sdk import Operations as op
+
+timestamp_format = "%Y-%m-%dT%H:%M:%SZ"
```
manjutapali: Constants should be in UPPERCASE

examples/key_based_replication/connector.py
```diff
@@ -0,0 +1,154 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# This shows key based replication from DB sources.
```
manjutapali, line 2: Add more details on what is and how key based replication works. It's better to add README file in this folder.

examples/key_based_replication/connector.py
```diff
@@ -0,0 +1,154 @@
...
+                            "email": row[3],  # Email id.
+                            "updated_at": dt2str(row[4])  # record updated at.
+                        })
+        # Storing `updated_at` of last fetched record
+        last_updated_at = dt2str(row[4])
+
+    # Update the state to the updated_at of the last record.
+    state["last_updated_at"] = last_updated_at
+    # Save the cursor for the next sync by yielding a checkpoint operation.
+    yield op.checkpoint(state)
```
manjutapali, line 119: We can recommend that, state can be saved with checkpoint after `N` records in case of large set of data.

examples/key_based_replication/connector.py
fivetran-rishabhghosh: Changed

examples/key_based_replication/connector.py
fivetran-rishabhghosh: Added in comments and not in readMe

shivangisingh172 (approved):

PR #31, opened 2024-09-16
feature(connector_sdk): Redshift DB example

fivetran-rishabhghosh:
Example to show how we can sync records from redshift DB via Connector SDK

### DIFF
--- a/examples/redshift/connector.py
+++ b/examples/redshift/connector.py
@@ -103,6 +103,7 @@ def update(configuration: dict, state: dict):
     last_updated_at = state["last_updated_at"] if "last_updated_at" in state else '2024-01-01T00:00:00Z'
 
     # Fetch records from DB sorted in ascending order.
+    # REPLICATION_KEY is updated_at column.
     query = (f"SELECT customer_id, first_name, last_name, email, updated_at FROM testers.customers WHERE {REPLICATION_KEY} > "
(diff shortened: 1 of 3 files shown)


examples/redshift/connector.py
```diff
@@ -0,0 +1,176 @@
...
+    cursor.execute("CREATE TABLE IF NOT EXISTS testers.customers "
+                   "(customer_id INTEGER NOT NULL, "
+                   "first_name VARCHAR, "
+                   "last_name VARCHAR, "
+                   "email VARCHAR, "
+                   "city VARCHAR, "
+                   "state VARCHAR, "
+                   "country VARCHAR, "
+                   "updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP)"
+                   " DISTSTYLE KEY DISTKEY (customer_id) SORTKEY (updated_at)")
```
shivangisingh172: I think we need to add a primary key constraint on `customer_id`. 
Also, a small description of `DISTSTYLE KEY` to describe usage.

ameetfivetran (approved):

examples/redshift/connector.py
fivetran-rishabhghosh: Added PK
Removed DISTSTYLE KEY

shivangisingh172 (approved):

PR #34, opened 2024-09-22
Feature(SDK): Add hash example

rajat-kushwaha:
Closes https://fivetran.height.app/T-764701

### Description of change
Adds an example showing how customer's can use a hash of all fields when there is no appropriate PK candidate in the source records.


examples/hashes/connector.py
```diff
@@ -0,0 +1,92 @@
...
+# See the technical reference documentation for more details on the update function
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update
+# The function takes two parameters:
+# - configuration: dictionary contains any secrets or payloads you configure when deploying the connector
+# - state: a dictionary contains whatever state you have chosen to checkpoint during the prior sync
+# The state dictionary is empty for the first sync or for any full re-sync
+def update(configuration: dict, state: dict):
+
+    # Represents a record fetched from source
+    row = {
```
varundhall: We need some more records here and also some interesting records where some of the fields, like email, are empty so that customers can get a better idea.
The whole idea of hash should be visible where customer cannot use composite PK.

varundhall (changes requested):

examples/hashes/connector.py
rajat-kushwaha: Added

examples/hashes/connector.py
```diff
@@ -0,0 +1,115 @@
+# This example shows how to calculate hash of fields to be used as primary key. This is useful in scenarios where the
+# incoming rows do not have any field suitable to be used as a Primary Key.
+# Fivetran recommends to define primary keys (https://fivetran.com/docs/connectors/connector-sdk/best-practices#declaringprimarykeys) when writing a connector.
+# This is to avoid disruptions caused due to a schema change (addition/removal of a column, etc) that might be needed in the maintainance phase of your connector.
```
varundhall:
```suggestion
# This is to avoid disruptions caused due to a schema change (addition/removal of a column, etc) that might be needed in the maintenance phase of your connector.
```

examples/hashes/connector.py
```diff
@@ -0,0 +1,115 @@
...
+# See the technical reference documentation for more details on the schema function
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#schema
+# The schema function takes one parameter:
+# - configuration: a dictionary that holds the configuration settings for the connector.
+# In this example, there is no use of the configuration
+def schema(configuration: dict):
+    return [
+        {
+            "table": "user",  # Name of the table in the destination.
+            "primary_key": ["_fivetran_id"],  # Primary key column(s) for the table.
```
varundhall: let's not use Fivetran id, customer should make their own column and not our system column

examples/hashes/connector.py
```diff
@@ -0,0 +1,115 @@
...
+# - state: a dictionary contains whatever state you have chosen to checkpoint during the prior sync
+# The state dictionary is empty for the first sync or for any full re-sync
+def update(configuration: dict, state: dict):
+
+    # Represents a record fetched from source
+    row_1 = {
+        "first_name": "John",  # First name.
+        "last_name": "Doe",  # Last name.
+        "email": "john.doe@example.com",  # Email ID.
+        "updated_at": "2007-12-03T10:15:30Z"  # Updated at timestamp.
```
varundhall: lets add two more rows
1 where everything is same but email is null
2 where all the values are same except updated_at

varundhall (changes requested):
Please reformat the file using PyCharm (Code->Reformat File)

examples/hashes/connector.py
```diff
@@ -0,0 +1,141 @@
+# This example shows how to calculate hash of fields to be used as primary key. This is useful in scenarios where the
```
varundhall: should we keep it here, or move to one for 3 category wise directories?

varundhall (approved):
Please link the example in README.md and also in our fivetran docs.

shivangisingh172 (approved):

examples/hashes/connector.py
rajat-kushwaha: Moved to `Common patterns for connector` folder

rajat-kushwaha:
> Please link the example in README.md and also in our fivetran docs.

Added in README file. 
Also, I found that in order to mention this in our [docs](https://fivetran.com/docs/connectors/connector-sdk/detailed-guide#examples), we'll need a link to code of this example. Hence I created this [ticket](https://fivetran.atlassian.net/browse/RD-801589) to update the docs once this PR gets merged.

PR #35, opened 2024-09-22
Feature(SDK): Add example to demonstrate how to deal with source records that do not have a created_at field

rajat-kushwaha:
Closes https://fivetran.height.app/T-764198

### Description of change
Adds an example showing how customer's can make `updated_at` (or equivalent) field as part of composite PK to obtain a history table-like behaviour.


examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,76 @@
...
+# See the technical reference documentation for more details on the update function
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update
+# The function takes two parameters:
+# - configuration: dictionary contains any secrets or payloads you configure when deploying the connector
+# - state: a dictionary contains whatever state you have chosen to checkpoint during the prior sync
+# The state dictionary is empty for the first sync or for any full re-sync
+def update(configuration: dict, state: dict):
+
+    # Represents a record fetched from source
+    row = {
```
varundhall: let's add 3 entries here
2 for normal record and 1 for an updated record

varundhall (changes requested):
please also update the PR title

examples/records_with_no_created_at_timestamp/connector.py
rajat-kushwaha: @varundhall I added one more record in code. The third I have put it in the comments section at the bottom of the file. Can you please check if that seems appropriate?

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,97 @@
...
+# Resulting table:
+# ┌───────────┬───────────────────┬────────────────────────────────────────────┬────────────────────────────┐
+# │    id     │   first_name      │     last_name     │    Designation         │         updated_at         │
+# │  integer  │     varchar       │      varchar      │      varchar           |  timestamp with time zone  │
+# ├───────────┼───────────────────┼───────────────────┼────────────────────────┤────────────────────────────│
+# │    123    │       John        │        Doe        │        Manager         │    2007-12-03T10:15:30Z    │
+# │    456    │       Jane        │       Dalton      │          VP            │    2007-12-03T10:15:30Z    │
+# └─────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+
+# Now lets say the record represented by row_1 gets updated in source as below:
```
varundhall: let's show this in code and not as comment.

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,97 @@
...
+# ┌───────────┬───────────────────┬───────────────────┬────────────────────────┬────────────────────────────┐
+# │    id     │   first_name      │     last_name     │    Designation         │         updated_at         │
+# │  integer  │     varchar       │      varchar      |      varchar           │  timestamp with time zone  │
+# ├───────────┼───────────────────┼───────────────────┤────────────────────────┼────────────────────────────│
+# │    123    │       John        │        Doe        │      Manager           │    2007-12-03T10:15:30Z    │
+# │    123    │       John        │        Doe        │   Senior Manager       │    2008-01-04T23:44:21Z    │
+# │    456    │       Jane        │       Dalton      │        VP              │    2007-12-03T10:15:30Z    │
+# └─────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+# Each sync will capture the latest update and will sync it as an additional record in the destination.
+# Updates between syncs cannot be captured this way. For explanation see https://fivetran.com/docs/core-concepts/sync-modes/history-mode#changestodatabetweensyncs
```
varundhall:
```suggestion
# Updates between syncs cannot be captured this way. For explanation see 
# https://fivetran.com/docs/core-concepts/sync-modes/history-mode#changestodatabetweensyncs
```

varundhall (changes requested):
Please rebase and reformat the file using PyCharm (Code->Reformat File)

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,91 @@
+# This is an example to show how to work with records where the source does not provide a created_at(or equivalent) field.
```
varundhall: do we need to keep this example here, or move it to one of the 3 folders?

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,91 @@
...
+# Refer to the Resulting table section at the bottom of this file for a detailed demonstration.
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
+
+# Import required classes from fivetran_connector_sdk
+from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Operations as op
+
+
+def schema(configuration: dict):
```
varundhall: add comment above schema method

varundhall (changes requested):
fivetran debug is throwing some error due to designation column, can you please check and ensure it works properly.

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,91 @@
...
+# - state: a dictionary contains whatever state you have chosen to checkpoint during the prior sync
+# The state dictionary is empty for the first sync or for any full re-sync
+def update(configuration: dict, state: dict):
+    # Represents a record fetched from source
+    row_1 = {
+        "id": 123,
+        "first_name": "John",  # First name.
+        "last_name": "Doe",  # Last name.
+        "designation": "Manager",  # Designation
+        "updated_at": "2007-12-03T10:15:30Z"  # Updated at timestamp.
```
fivetran-satvikpatil: I am not understanding this part clearly, are you assuming that we will be getting this `updated_at` column from the source? If that is the case, then yes having it as part of PK should solve the problem. 

But what if the customer table doesn't have any columns like `created_at` and `updated_at`, should we also cover this scenario, what customer can do in this case? 
E.g. we can add the `updated_at` column from the code (the timestamp would be during sync_time) and this would also help them identify which row was synced at what time.

examples/records_with_no_created_at_timestamp/connector.py
rajat-kushwaha: Moved to `Common patterns for connector` folder

examples/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,91 @@
...
+    row_1_updated = {
+        "id": 123,
+        "first_name": "John",
+        "last_name": "Doe",
+        "designation": "Senior Manager",  # Value changed
+        "updated_at": "2008-01-04T23:44:21Z"  # Updated at changed
+    }
+
+    # Represents another record fetched from source
+    row_2 = {
```
fivetran-satvikpatil: Also, should we use any free api to replicate this behavior instead of hardcoding records?

examples/records_with_no_created_at_timestamp/connector.py
rajat-kushwaha: > are you assuming that we will be getting this updated_at column from the source
Yes, the assumption here is that the source will provide an `updated_at` (or equivalent - `modified_at` etc) column. 

`But what if the customer table doesn't have any columns like created_at and updated_at`
It was decided that we do not want to handle this case through code.  IMO the `_fivetran_synced` system column should suffice in this scenario. 

`we can add the updated_at column from the code (the timestamp would be during sync_time) and this would also help them identify which row was synced at what time.`
It was discussed that we should not be doing this as this ain't good practice and the `_fivetran_synced` system column should suffice.

fivetran-satvikpatil: Ohhk, got it.

examples/records_with_no_created_at_timestamp/connector.py
rajat-kushwaha: I think we should keep the examples simple. Hard coding the examples would clearly reflect what changed in the source record.

fivetran-satvikpatil (approved):

examples/Common patterns for connectors/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,96 @@
+# This is an example to show how to work with records where the source does not provide a created_at(or equivalent) field.
+# If you want to keep track of when the record was first observed, then the updated_at(or equivalent, eg, modified_at etc)
+# field can be made part of the composite Primary Key. By doing this you can ensure that successive syncs do not rewrite the
+# updated_at field in the existing row in the destination. Instead, it creates a new row with the new updated_at.
+# Now, the record with earliest value of _fivetran_synced system column represents when the record was first observed in source.
+# Refer to the Resulting table section at the bottom of this file for a detailed demonstration.
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
```
varundhall:
```suggestion
# This example demonstrates how to handle records when the source does not provide a created_at (or equivalent) field.
# If you want to track when a record was first observed, the updated_at (or an equivalent, e.g., modified_at) field can
# be made part of the composite primary key. This ensures that successive syncs do not overwrite the updated_at field in
# the existing row in the destination. Instead, a new row is created with the new updated_at value.
# The record with the earliest value of the _fivetran_synced system column will indicate when the record was first
# observed in the source.
# Refer to the "Resulting Table" section at the bottom of this file for a detailed demonstration.
# See the Technical Reference documentation
# (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
```

examples/Common patterns for connectors/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Import required classes from fivetran_connector_sdk
+from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Operations as op
+
+# Define the schema function which lets you configure the schema your connector delivers.
```
varundhall:
```suggestion

# Define the schema function which lets you configure the schema your connector delivers.
```

varundhall: please reformat your file, using Code->Reformat File in PyCharm

examples/Common patterns for connectors/records_with_no_created_at_timestamp/connector.py
```diff
@@ -0,0 +1,96 @@
...
+    return [
+        {
+            "table": "user",  # Name of the table in the destination.
+            "primary_key": ["id", "updated_at"],  # Primary key column(s) for the table.
+            "columns": {  # Define the columns and their data types.
+                "id": "INT",
+                "updated_at": "UTC_DATETIME",  # UTC date-time column for the updated_at.
+                "first_name": "STRING",  # String column for the first name.
+                "last_name": "STRING",  # String column for the last name.
+                "designation": "STRING"
```
varundhall: please be consistent with comments

varundhall (changes requested):
please rebase and fix your directory structure

varundhall (approved):

PR #44, opened 2024-10-20
fix(User Profiles Example): Handle NaN value from dataframes

fivetran-satvikpatil:
T-802267

### Changes
Provide example usage to remove the NaN value from the data frame before upserting.

### Testing
- `fivetran debug`
Screenshot 2024-10-23 at 12 47 53 PM


examples/quick_start_examples/user_profiles/connector.py
```diff
@@ -9,6 +9,7 @@
 # Import the requests module for making HTTP requests, aliased as rq.
 import requests as rq
 import pandas as pd
+import numpy as np
```
varundhall: please add in requirements

examples/quick_start_examples/user_profiles/connector.py
```diff
@@ -86,6 +91,42 @@ def update(configuration: dict, state: dict):
...
+    # Upsert tables with NaN
+    cursor_2 = state["table_with_nan_cursor"] if "table_with_nan_cursor" in state else 0
+    table_df_1 = generate_data_with_NaN()
+    table_df_2 = generate_data_with_NaN()
+    table_df_3 = generate_data_with_NaN()
+
+    # Approaches to replace NaN with None
+
+    # Convert NaN to None using the replace method
+    table_df_1 = table_df_1.replace(np.nan, None)
```
varundhall: I think we should create a method and add proper comments like
Approach 1:
Approach 2:
etc

varundhall (changes requested):

varundhall (approved):

examples/quickstart_examples/user_profiles/connector.py
```diff
@@ -53,38 +58,14 @@ def update(configuration: dict, state: dict):
...
-    # UPSERT all login table data.
-    # Iterate over the values of the dictionary (which are the DataFrame rows)
-    for value in login_df.to_dict("index").values():
-        yield op.upsert("login", value)
+    yield from upsert_dataframe_approach_3(login_df=login_df, state=state, cursor=cursor)
 
-    # Checkpointing at the end of the "login" table data processing
-    state["login_cursor"] = cursor
-    yield op.checkpoint(state)
+    # Approaches to handle NaN values in dataframes
```
fivetran-sadiemartin, line 70: Let's explain why NaN presents a problem here.

fivetran-sadiemartin (changes requested):
Could we add to the comment at the top that this example shows how to work with data frames and how to handle NaN in data frames?

I didn't run the example - I'm assuming someone on engineering did to make sure that it still works?

PR #46, opened 2024-10-28
fix(examples): add example logs and fix naming

varundhall:
Closes https://fivetran.height.app/T-805250

### DIFF
--- a/examples/quickstart_examples/simple_three_step_cursor/connector.py
+++ b/examples/quickstart_examples/simple_three_step_cursor/connector.py
@@ -1,5 +1,5 @@
 # This is a simple example for how to work with the fivetran_connector_sdk module.
-# It uses a local SOURCE_DATA object created at the start of the script, and demonstrates the Schema() method use.
+# It uses a simple_three_step_cursor SOURCE_DATA object created at the start of the script, and demonstrates the Schema() method use.
 # It also shows a way to manage state.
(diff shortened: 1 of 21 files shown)


examples/quickstart_examples/simple_three_step_cursor/connector.py
```diff
@@ -42,12 +42,16 @@ def schema(configuration: dict):
...
 def update(configuration: dict, state: dict):
+    log.warning("Example: QuickStart Examples - Simple Three Step Cursor")
+
     # Retrieve the cursor from the state object to determine the current position in the SOURCE_DATA.
     # If the cursor is not present in the state, start from the beginning (cursor = 0).
     cursor = state['cursor'] if 'cursor' in state else 0
     log.fine(f"current cursor is {repr(cursor)}")
 
     # Get the row of data from SOURCE_DATA using the cursor position.
+    if cursor >= SOURCE_DATA.__len__():
```
fivetran-satvikpatil, line 53: Do we want to throw an exception? Otherwise, we can take the modulus of the cursor by 3 and return the related data at that index.

fivetran-satvikpatil (approved):

examples/quickstart_examples/simple_three_step_cursor/connector.py
varundhall: we want the connector to break and stop after 3 iterations, otherwise it will run forever.

alsnkline (commented):
Love it

PR #50, opened 2024-11-06
Feature(ConnectorSdk): Accelo Api Connector Multithreading Enabled Example

fivetran-chiragramachandraiah:
Closes https://fivetran.height.app/T-809939

Description:
Accelo source connector example with multithreading support.

**Testing:**
* Testing pending on source access, will update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,13 @@ You would need to provide your redshift credentials for this example to work.
 This shows key based replication from DB sources.
 Replication keys are columns that are used to identify new and updated data for replication.
 When you set a table to use Incremental Replication, you’ll also need to define a replication key for that table.
+
+### accelo api connector multithreading enabled
+This module implements a connector for syncing data from the Accelo API to Fivetran.
+It handles authentication, rate limiting, and data synchronization for companies,
+invoices, payments, prospects, jobs, and staff.
+Multithreading is used to make api calls in parallel to pull data faster.
+Requires Accelo credentials to be passed in to work.
 </details>
 
 # Additional considerations
--- a/examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
+++ b/examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
@@ -1,0 +1,828 @@
+"""
+Accelo API Connector for Fivetran
+
+This module implements a connector for syncing data from the Accelo API to Fivetran.
+It handles authentication, rate limiting, and data synchronization for companies,
+invoices, payments, prospects, jobs, and staff.
+
+Author: Ahmed Zedan
+Date: 2024-09-20
+Version: 1.3
+"""
+
+from fivetran_connector_sdk import Connector, Logging as log, Operations as op
+import requests
+import time
+from datetime import datetime, timezone
+from concurrent.futures import ThreadPoolExecutor
+from threading import local
+from copy import deepcopy
+
+from dotenv import load_dotenv
+
+# Load environment variables from .env file
+load_dotenv()
+
+# Set the log level explicitly
+log.LOG_LEVEL = log.Level.INFO
+
+# Constants
+BASE_URL = "https://{deployment}.api.accelo.com/api/v0"
+RATE_LIMIT = 5000  # Requests per hour
+RATE_LIMIT_INTERVAL = 3600  # 1 hour in seconds
+CHECKPOINT_INTERVAL = 1000  # Number of records to process before creating a checkpoint
+MAX_WORKERS = 5  # Number of concurrent API requests
+REQUEST_TIMEOUT = 30  # Default timeout in seconds
+SYNC_TIMEOUT = 3600  # Timeout for the entire sync operation in seconds
+BATCH_SIZE = 100  # Maximum number of records per page in Accelo API (API limit)
+RETRIES = 3
+
+# Create a thread-local state object
+thread_local_state = local()
+
+
+def get_access_token(client_id, client_secret, deployment):
+    """
+    Obtain an access token from the Accelo API using OAuth 2.0 client credentials flow.
+
+    Args:
+        client_id (str): The OAuth 2.0 client ID.
+        client_secret (str): The OAuth 2.0 client secret.
+        deployment (str): The Accelo deployment name.
+
+    Returns:
+        str: The access token if successful.
+
+    Raises:
+        Exception: If authentication fails.
+    """
+    log.info("Obtaining access token")
+    token_url = f"https://{deployment}.api.accelo.com/oauth2/v0/token"
+    data = {
+        "grant_type": "client_credentials",
+        "client_id": client_id,
+        "client_secret": client_secret,
+        "scope": "read(all)"
+    }
+    response = requests.post(token_url, data=data)
+    if response.status_code == 200:
+        log.info("Access token obtained successfully")
+        return response.json()["access_token"]
+    else:
+        log.severe(f"Failed to obtain access token: {response.text}")
+        raise Exception("Failed to obtain access token")
+
+
+def fetch_data(endpoint, access_token, params=None, timeout=REQUEST_TIMEOUT, retries=RETRIES):
+    """
+    Fetch data from the Accelo API with retry logic.
+
+    Args:
+        endpoint (str): The API endpoint to fetch data from.
+        access_token (str): The OAuth 2.0 access token.
+        params (dict, optional): Query parameters for the API request.
+        timeout (int, optional): Request timeout in seconds.
+        retries (int, optional): Number of retry attempts.
+
+    Returns:
+        list: The JSON response data or None if all retries fail.
+    """
+    url = f"{BASE_URL}/{endpoint}"
+    headers = {"Authorization": f"Bearer {access_token}"}
+
+    for attempt in range(retries):
+        try:
+            response = requests.get(url, headers=headers, params=params, timeout=timeout)
+            response.raise_for_status()
+            return response.json().get("response", [])
+        except requests.exceptions.RequestException as e:
+            log.warning(f"Error fetching data from {endpoint} (attempt {attempt + 1}/{retries}): {str(e)}")
+            if attempt == retries - 1:
+                log.severe(f"Failed to fetch data from {endpoint} after {retries} retries")
+                return None
+            time.sleep(2 ** attempt)  # Exponential backoff
+
+
+def update(configuration: dict, state: dict):
+    """
+    Main update function for the connector.
+
+    This function orchestrates the entire sync process:
+    1. Authenticates with the Accelo API
+    2. Syncs data for companies, invoices, payments, prospects, jobs, and staff
+    3. Yields checkpoints to update the sync state
+
+    Args:
+        configuration (dict): The connector configuration.
+        state (dict): The current state of the connector.
+
+    Yields:
+        dict: Checkpoint operations to update the sync state.
+    """
+    log.info(f"Starting update process. Initial state: {state}")
+    thread_local_state.state = deepcopy(state)
+
+    client_id = configuration.get("client_id")
+    client_secret = configuration.get("client_secret")
+    deployment = configuration.get("deployment")
+
+    if not all([client_id, client_secret, deployment]):
+        log.severe("Missing required configuration parameters")
+        return
+
+    try:
+        access_token = get_access_token(client_id, client_secret, deployment)
+        global BASE_URL
+        BASE_URL = f"https://{deployment}.api.accelo.com/api/v0"
+
+        update_start_time = time.time()
+
+        for entity_sync in [sync_companies, sync_invoices, sync_payments, sync_prospects, sync_jobs, sync_staff]:
+            entity_name = entity_sync.__name__.replace("sync_", "")
+            yield from entity_sync(access_token)
+
+        yield op.checkpoint(thread_local_state.state)
+
+        update_duration = time.time() - update_start_time
+        log.info(
+            f"Update process completed. Total time: {update_duration:.2f} seconds. Final state: {thread_local_state.state}")
+    except Exception as e:
+        log.severe(f"Update process failed: {str(e)}")
+
+
+def sync_entity(entity_name, access_token, fields, last_sync_key, process_record=None, date_field="date_modified",
+                timeout=SYNC_TIMEOUT, batch_size=BATCH_SIZE, fetch_data_func=None):
+    """
+    Sync data for a specific entity from Accelo API.
+
+    This function handles the core logic for syncing each entity type:
+    - Fetches data in batches
+    - Processes records
+    - Yields upsert operations
+    - Updates the sync state
+
+    Args:
+        entity_name (str): The name of the entity being synced.
+        access_token (str): The OAuth 2.0 access token.
+        fields (list): The fields to sync for this entity.
+        last_sync_key (str): The key to use for storing the last sync time in the state.
+        process_record (callable, optional): A function to process each record before upserting.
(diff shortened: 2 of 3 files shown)


examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,828 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from threading import local
+from copy import deepcopy
+
+from dotenv import load_dotenv
+
+# Load environment variables from .env file
+load_dotenv()
+
+# Set the log level explicitly
```
fivetran-satvikpatil: This is not actually required, we initialise it in `__init__.py`.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,828 @@
...
+                "status": "INT"
+            }
+        }
+    ]
+    return schema_definition
+
+
+connector = Connector(update=update, schema=schema)
+
+if __name__ == "__main__":
```
fivetran-satvikpatil, line 696: We do need `configuration.json` file here. We need to change this to work in debug mode (check marketstack example).

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,828 @@
...
+
+    try:
+        access_token = get_access_token(client_id, client_secret, deployment)
+        global BASE_URL
+        BASE_URL = f"https://{deployment}.api.accelo.com/api/v0"
+
+        update_start_time = time.time()
+
+        for entity_sync in [sync_companies, sync_invoices, sync_payments, sync_prospects, sync_jobs, sync_staff]:
+            entity_name = entity_sync.__name__.replace("sync_", "")
```
fivetran-satvikpatil: Where/how are we using this variable `entity_name`?

README.md
```diff
@@ -98,6 +98,13 @@ You would need to provide your redshift credentials for this example to work.
 This shows key based replication from DB sources.
 Replication keys are columns that are used to identify new and updated data for replication.
 When you set a table to use Incremental Replication, you’ll also need to define a replication key for that table.
+
+### accelo api connector multithreading enabled
+This module implements a connector for syncing data from the Accelo API to Fivetran.
+It handles authentication, rate limiting, and data synchronization for companies,
```
fivetran-satvikpatil: Lets also mention it uses OAuth2, as we needed an example for this anyway.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,828 @@
...
+        entities_received = 0
+        start_time = time.time()
+        max_date_value = last_sync_unix
+
+        def fetch_page(page):
+            nonlocal entities_received, max_date_value
+            current_params = params.copy()
+            current_params['_page'] = page
+            try:
+                entities = fetch_data_func(current_params) if fetch_data_func else fetch_data(entity_name, access_token,
```
fivetran-satvikpatil, line 201: Is there any place we provide `fetch_data_func`, I think we can directly use `fetch_data`, please confirm and remove it.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,828 @@
...
+        for field in int_fields:
+            value = record.get(field)
+            if value is not None:
+                try:
+                    record[field] = int(value)
+                except (ValueError, TypeError):
+                    log.warning(f"Could not convert field {field} with value '{value}' to int")
+                    record[field] = None
+
+        # Convert date fields
```
fivetran-satvikpatil: This logic of converting date fields could be a common method. It is same in each method.
`sync_companies`, `sync_payments`, `sync_prospects`, ... etc

Same for integer fields as well.

fivetran-satvikpatil (commented):
Please add the `duckdb` and `fivetran debug` screenshots in the PR description.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Yes it's not used anywhere else, but I think its to keep it customisable, not sure so I think we can let it be as it's minor.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Done

README.md
fivetran-chiragramachandraiah: Done

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Removed.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Done

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Done.

fivetran-chiragramachandraiah:
> Please add the duckdb and fivetran debug screenshots in the PR description.

Final test done. Screenshots added.

README.md
```diff
@@ -98,6 +98,13 @@ You would need to provide your redshift credentials for this example to work.
 This shows key based replication from DB sources.
 Replication keys are columns that are used to identify new and updated data for replication.
 When you set a table to use Incremental Replication, you’ll also need to define a replication key for that table.
+
+### accelo api connector multithreading enabled
+This module implements a connector for syncing data from the Accelo API to Fivetran.
```
fivetran-sadiemartin: remove "to Fivetran"

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,760 @@
+"""
+Accelo API Connector for Fivetran
+
+This module implements a connector for syncing data from the Accelo API to Fivetran.
+It handles authentication, rate limiting, and data synchronization for companies,
+invoices, payments, prospects, jobs, and staff.
```
fivetran-sadiemartin, line 6: Let's add here that it is an example of how to use multithreading to improve connector performance.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,760 @@
...
+
+Author: Ahmed Zedan
```
fivetran-sadiemartin: @alsnkline what should we do about attribution? I was thinking we should leave it off unless we get specific permission from Ahmed, but this is such a detailed and thorough example, I don't want to pass it off as our own. Are you okay with just putting his name here like this?

alsnkline: I am ok with his name as it will make him proud and there are many Ahemd Zedan's
in the world - the only think odd is that its not clear they are a customer rather than one of our team members.

I was thinking we should make more of a standard 'attribution' sentence.
maybe 
"Example submitted by our amazing community member Ahmed Zedan"
"Example submitted by community member Ahmed Zedan"

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,760 @@
...
+Date: 2024-09-20
+Version: 1.3
```
alsnkline: What is the version for?

if its related to the example we should remove it - its not our 1.3 it would be Ahmed's
if its related to a version of Accelo API then we should keep it but make that clear.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
alsnkline: Multithreading the extraction of data from the source - we need to be clear about what is being mutithreaded and its not fivetran or our loading of your data.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,760 @@
...
+BATCH_SIZE = 100  # Maximum number of records per page in Accelo API (API limit)
+RETRIES = 3
+
+# Create a thread-local state object
+thread_local_state = local()
+
+
+def get_access_token(client_id, client_secret, deployment):
+    """
+    Obtain an access token from the Accelo API using OAuth 2.0 client credentials flow.
```
alsnkline, line 46: And we also have an example on using OAuth to get data from a source.

we should just reference this example rather than duplicate - DRY.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
alsnkline: We also need to call out that this is also an example of using OAuth 2.0 client credentials flow

alsnkline (approved):
looking good - couple of suggestions

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: @fivetran-sadiemartin @alsnkline I have added similar points just below. I will rephrase it.

> It handles OAuth2 authentication, rate limiting, and data synchronization for companies,
> invoices, payments, prospects, jobs, and staff.
> Multithreading is used to make api calls in parallel to pull data faster.
> Requires Accelo credentials to be passed in to work.

fivetran-chiragramachandraiah: I had added similar points in the description. I have rephrased it and copied it over.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Done

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Removed.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: I don't see any OAuth e.g. currently..

examples/source_examples/accelo_api_connector_multithreading_enabled/configuration.json
```diff
@@ -0,0 +1,5 @@
+{
+  "deployment": "<DEPLOYMENT>",
+  "client_id": "<CLIENT_ID>>",
```
fivetran-satvikpatil: Replace <CLIENT_ID> with <YOUR_CLIENT_ID>
We follow similar pattern for all the configuration.json

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,763 @@
...
+    Sync job data from Accelo API.
+    """
+    job_fields = [
+        "id", "title", "status", "standing", "date_created", "date_started",
+        "date_due", "date_completed", "date_modified", "manager", "company",
+        "contact", "description", "value"
+    ]
+
+    def process_job_record(record):
+        # Convert integer fields
```
fivetran-satvikpatil: You can create a common method for converting integer and float fields as well.

fivetran-satvikpatil (approved):
Added minor comments, please do the changes and test it before merging, else LGTM.

examples/source_examples/accelo_api_connector_multithreading_enabled/configuration.json
fivetran-chiragramachandraiah: Done.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
fivetran-chiragramachandraiah: Done.

shivangisingh172 (commented):
As discussed earlier, multithreaded approach for source needs to ensure thread safety for core operations (upsert/update/delete/checkpoint) and should be synchronised. 

We must highlight that in the example as well.

examples/source_examples/accelo_api_connector_multithreading_enabled/connector.py
```diff
@@ -0,0 +1,701 @@
...
+                        if entity_date != 'unknown':
+                            try:
+                                if isinstance(entity_date, str) and 'T' in entity_date:
+                                    dt = datetime.fromisoformat(entity_date.replace('Z', '+00:00'))
+                                else:
+                                    timestamp = int(entity_date)
+                                    dt = datetime.fromtimestamp(timestamp, tz=timezone.utc)
+                                date_value_unix = int(dt.timestamp())
+                                if date_value_unix > max_date_value:
+                                    max_date_value = date_value_unix
```
fivetran-chiragramachandraiah, line 218: Thanks for pointing out @shivangisingh172. In this e.g. the multithreading happens over the `fetch_page` method. And no core operations are called in it so it's safe. We need to include you're point as a multithreading guideline. Will do it. Guideline can be:
* Never call any Operations inside multithreaded code.

On a side note I found 1 possible issue. The multithreading happens over `fetch_page` method. In that method a common field `max_date_value` is used which is used to track cursor progress. It is possible a race condition can happen over it. Like it won't fail but it may be wrongly set. So to mitigate it, I'm thinking below:
* In the multithreading `fetch_page` method just purely make the api call. Remaining code logic move it outside to the synchronous code path so that no race conditions can happen.

I'll make the above change.
Also the guideline in this case can be:
* Use multithreading purely to make api calls and get the response, remaining logic keep it single-threaded/synchronous to avoid race conditions.

fivetran-chiragramachandraiah: Made the changes. Pls have a look.

fivetran-chiragramachandraiah: @fivetran-sadiemartin @alsnkline I've added the multithreading guidelines in api_threading_utils.py. Also updated the e.g. descriptions. Pls review the phrasing.

examples/source_examples/accelo_api_connector_multithreading_enabled/api_threading_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+executor = ThreadPoolExecutor(max_workers=constants.MAX_WORKERS)
+
+# Util function to make api calls in parallel using threadPoolExecutor
+def make_api_calls_in_parallel(page, fetch_page):
+    futures = [executor.submit(fetch_page, p) for p in range(page, page + constants.MAX_WORKERS)]
+    results = [future.result() for future in futures]
+    return results
+
+# Util funtion to make an api call and return response
```
alsnkline:
```suggestion
# Util function to make an api call and return response
```

alsnkline (approved):

PR #101, opened 2025-03-28
feature(examples): add way to provide location for weather example and fix comment

varundhall:
Closes https://fivetran.height.app/T-913363

Updated Weather example to make forecast location configurable via configuration.json with default values as fallback.

### Testing
Without any `configuration.json`
image

With `configuration.json`
image

### DIFF
--- a/examples/quickstart_examples/weather/connector.py
+++ b/examples/quickstart_examples/weather/connector.py
@@ -1,5 +1,5 @@
 # This is a simple example for how to work with the fivetran_connector_sdk module.
-# It shows a connector that calls a publicly available API to get the weather forecast data for Myrtle Beach in South Carolina, USA.
+# It shows a simple connector that calls a publicly available API to get the weather forecast data for the defined location.
 # It also shows how to use the logging functionality provided by fivetran_connector_sdk, by logging important steps using log.info() and log.fine()
 # See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
 # and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details.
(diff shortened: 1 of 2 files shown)


examples/quickstart_examples/weather/connector.py
```diff
@@ -52,8 +52,15 @@ def update(configuration: dict, state: dict):
...
 
-    # Get weather forecast for Myrtle Beach, SC from National Weather Service API.
-    response = rq.get("https://api.weather.gov/gridpoints/ILM/58,40/forecast")
+    # Read values from configuration with defaults weather forecast for Myrtle Beach, SC from National Weather Service API.
+    forecast_office_id = configuration.get('forecast_office_id', 'ILM')
+    x_coordinate = configuration.get('x_coordinate', 58)
+    y_coordinate = configuration.get('y_coordinate', 40)
+
+    # Make the API request using the values
+    log.info(f"Getting forecast for Office Id: {forecast_office_id} with x_coordinate: {x_coordinate} and y_coordinate: {y_coordinate}")
```
fivetran-rishabhghosh: I would refrain from logging this.
Reasoning: Customers might have many coordinates they are reading data from. They might start logging every request in this case. Is this not something we should avoid?

examples/quickstart_examples/weather/connector.py
```diff
@@ -99,8 +106,15 @@ def update(configuration: dict, state: dict):
 # This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
 # Please test using the Fivetran debug command prior to finalizing and deploying your connector.
 if __name__ == "__main__":
-    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE.
-    connector.debug()
+    try:
+        # Try loading the configuration from the file
+        with open("configuration.json", 'r') as f:
+            configuration = json.load(f)
+    except FileNotFoundError:
```
fivetran-rishabhghosh, line 113: This exception handling is not added for other examples. Should we create a task to add them ?

fivetran-rishabhghosh (approved):
Please check if you would like to remove the above log

examples/quickstart_examples/weather/connector.py
varundhall: Yeah we can avoid it

fivetran-sahilkhirwal (approved):
LGTM

PR #106, opened 2025-04-07
fix(examples): update state usage example

alsnkline:
Making the timestamps for the second department unique so the state management is clearer.

### DIFF
--- a/examples/common_patterns_for_connectors/cursors/multiple_tables_with_cursors/connector.py
+++ b/examples/common_patterns_for_connectors/cursors/multiple_tables_with_cursors/connector.py
@@ -86,7 +86,7 @@ def update(configuration: dict, state: dict):
 
     # State at the end of sync for simulated records
     # {"company_cursor": "2024-08-14T02:01:00Z",
-    # "department_cursor": {"1": "2024-08-14T03:00:00Z", "2": "2024-08-14T03:00:00Z"}}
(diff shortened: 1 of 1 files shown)


alsnkline:
@varundhall now I'm playing with this example I'm not sure that it is really demonstrating how to use state.

right now the primary keys are such that company 2's departments over write company 1's departments.
that's weird.

also the cursor isn't used when getting data so it doesn't show using the cursor only setting it with check point.

I think it would be more powerful to set company id as part of PK for department table so we end up with 6 records there.
1 for each department.

and set the updated_at as different days - so it simulates new departments getting setup in these companies.

it would good if the first sync got 4 results - all with updated_at before some time stamp and then the second sync used that time stamp to get the rest - demonstrating how using the cursor in the API call gets a different set of records.

Am I missing something about how we were expecting this to be used?

alsnkline:
Would it actually be easier to move these examples of using state to the playground?
then you could use the first sync to capture some number of base records and they create a random number of new 'updated' records each sync and use the current time as the cursor.

varundhall:
@alsnkline, it looks like there is a bug in the connector; I will fix it.
Also, I will create a follow-up ticket to move this to the playground.

varundhall:
Tickets: 
- https://fivetran.atlassian.net/browse/RD-923692
- https://fivetran.atlassian.net/browse/RD-923693

varundhall (approved):

PR #113, opened 2025-04-09
task(connector-sdk): Adding the GPG private key example

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-868132

### What the connector does:
- Fetches the required private key and passphrase from configuration
- Loads it in the GPG object and signs a message
- Upserts the signed message

### Verification:
Screenshot 2025-04-09 at 17 28 25

### DIFF
--- a/examples/common_patterns_for_connectors/gpg_private_keys/connector.py
+++ b/examples/common_patterns_for_connectors/gpg_private_keys/connector.py
@@ -1,0 +1,149 @@
+# This is a simple example for how to work with the fivetran_connector_sdk module.
+# It defines a simple `update` method, which uses python-gnupg to sign messages with a private key.
+# This example upserts the signed data to a table named "signed_message".
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
+
+# Import required classes from fivetran_connector_sdk
+from fivetran_connector_sdk import Connector # For supporting Connector operations like Update() and Schema()
+from fivetran_connector_sdk import Logging as log # For enabling Logs in your connector code
+from fivetran_connector_sdk import Operations as op # For supporting Data operations like Upsert(), Update(), Delete() and checkpoint()
+
+# Importing the gnupg library for GPG encryption and decryption
+import gnupg
+import os
+import json
+
+
+# This method is used to set up the GPG home directory.
+# This is required for the gnupg library to work properly.
+# The GPG home directory is where the GPG keys and configuration files are stored.
+# This method returns the path to the GPG home directory.
+def setup_gnupghome():
+    # Set up the GPG home directory
+    # You can change this to a different directory if you want
+    gnupghome = './gnupg'
+    if not os.path.exists(gnupghome):
+        log.info(f"Creating GPG home directory at {gnupghome}")
+        os.makedirs(gnupghome)
+    return gnupghome
+
+
+# This method is used to import the private key into the GPG object.
+# The private key is used to sign messages.
+# The method takes the configuration dictionary and the GPG object as parameters.
+# The method returns the imported key.
+def get_gpg_key(configuration: dict, gpg):
+    # Get the private key from the configuration
+    # Ensure that the private key string in configuration is properly escaped as JSON string.
+    private_key = configuration.get("private_key")
+    key = gpg.import_keys(private_key)
+    # Check if the key was imported successfully
+    if key.count == 0:
+        raise RuntimeError("Failed to import the private key.")
+    else:
+        log.info("Successfully imported the private key.")
+        return key
+
+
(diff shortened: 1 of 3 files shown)


examples/common_patterns_for_connectors/gpg_private_keys/connector.py
```diff
@@ -0,0 +1,140 @@
...
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
+if __name__ == "__main__":
+    try:
+        # Try loading the configuration from the file
+        with open("configuration.json", 'r') as f:
+            configuration = json.load(f)
+    except FileNotFoundError:
+        # Fallback to an empty configuration if the file is not found
+        configuration = {}
```
fivetran-rishabhghosh: Is this fallback required? Would the example not fail if configurations is not passed?

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # This is the message to be signed
+    # You can change this to any message you want to sign
+    message = "This is a test message. This will be encrypted and signed"
+
+    # Sign the message using the GPG object and the imported key
+    signed_message = sign_message(configuration, gpg, key, message)
+
+    # Verify if the signed message is valid
+    verified = gpg.verify(signed_message)
```
fivetran-rishabhghosh, line 107: gpg.verify(signed_message) does not return a simple boolean. Can you please cehck

fivetran-rishabhghosh (changes requested):
Requested configuration.json on slack to verify.
Please review my comments and I will approve once I validate

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
fivetran-sahilkhirwal: I referred to the weather example. It had this fallback, so i added it.
Ideally there is another check in the schema to ensure that the configuration has the required values for the example. So in the case when the configuration is not passed, it will throw an error stating that the required key-value pairs are missing.

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
fivetran-sahilkhirwal: gpg.verify() returns an object of type `Verify`.
The gnupg.Verify object implements the __bool__() method, which allows it to be used in boolean contexts like if statements.
So in case of if blocks, the Verify object returns the value of `Verify.valid` which is a boolean.
Reference from the documentation:
Screenshot 2025-04-11 at 00 14 06

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
fivetran-rishabhghosh: In the weather example, the fallback is present because it has default values. Here that is not the case, you can remove this

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
```diff
@@ -0,0 +1,138 @@
...
+import gnupg
+import os
+import json
+
+
+# This method is used to set up the GPG home directory.
+# You can change the GPG home directory to a different location if you want.
+# The GPG home directory is where the GPG keys and configuration files are stored.
+# This method returns the path to the GPG home directory.
+def setup_gnupghome():
```
fivetran-rishabhghosh: As discussed let's remove this

fivetran-rishabhghosh (changes requested):

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
fivetran-sahilkhirwal: done!

fivetran-rishabhghosh (approved):

examples/common_patterns_for_connectors/gpg_private_keys/connector.py
```diff
@@ -0,0 +1,127 @@
...
+        return key
+
+
+# This method is used to sign a message using the GPG object.
+# The method takes the configuration dictionary, the GPG object, the key, and the message to be signed as parameters.
+# The method returns the signed message.
+def sign_message(configuration, gpg, key, message):
+    passphrase = configuration.get("passphrase")
+    # Extract the key ID from the imported key and sign the message using the key ID and passphrase
+    key_id = key.fingerprints[0]
```
fivetran-anushkaparashar, line 54: Can we mention how to identify and extract the key to use in case of multiple keys?

fivetran-pranavtotala: Do add documentation link on how to generate private keys on your machine.

fivetran-sahilkhirwal: Addressed both the comments

PR #131, opened 2025-05-01
Added examples/common_patterns_for_connectors/tracking_tables

fivetran-clgritton:
Added an example for tracking in state which tables have been synced. Any new tables added to the connector.py that haven't been synced will have a historical sync, others will sync using the timestamp of last sync.

image
image


examples/common_patterns_for_connectors/tracking_tables/connector.py
```diff
@@ -0,0 +1,118 @@
...
+    yield from sync_tables(tables_previously_synced, from_timestamp, initial_timestamp)
+
+    """
+    Save the progress by checkpointing the state. This stores the list of tables that are up-to-date.
+    Learn more about how and where to checkpoint by reading our best practices documentation
+    (https://fivetran.com/docs/connectors/connector-sdk/best-practices#largedatasetrecommendation).
+    """
+    tables_synced_this_sync.sort()
+    new_state["synced_tables"] = tables_synced_this_sync
+    yield op.checkpoint(new_state)
```
alsnkline, line 70: Should we have a fine log statement here printing out new_state as it gets checkpoint?

examples/common_patterns_for_connectors/tracking_tables/connector.py
```diff
@@ -0,0 +1,118 @@
...
+# This example does not use the schema() function. If it did, it would need to be included in the connector object definition.
+connector = Connector(update=update)
+
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
+if __name__ == "__main__":
+    # Adding this code to your `connector.py` allows you to test your connector by running your file directly from your IDE:
+    connector.debug()
```
alsnkline, line 119: wont this need configuration to work?

I don't see a configuration.json file in the example => shouldn't we provide one as you use it?

Although you also have a constants that seems to be logged - is it used?

alsnkline (approved):
a few thoughts - love it generally - thank you for creating

examples/common_patterns_for_connectors/tracking_tables/connector.py
fivetran-clgritton: great idea, will add.

examples/common_patterns_for_connectors/tracking_tables/connector.py
fivetran-clgritton: adding configuration.json

PR #136, opened 2025-05-05
added azure keyvault example

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-938443

### What the connector does:
- It fetches the secrets stored in azure keyvault using the azure credentials
- It uses the fetched secrets to connect to postgresql database
- executes a query on postgresql to verify the connection is correct
- Upsert the record to the destination

### Validation 

Screenshot 2025-05-05 at 14 54 38
Screenshot 2025-05-05 at 14 54 48

### DIFF
--- a/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
+++ b/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
@@ -73,11 +73,7 @@ The connector retrieves credentials from Azure Key Vault and uses them to establ
 
 1. Executes a simple query to verify connectivity and retrieve data from a table in the database.
 
-2. Upserts the data in the `employees` table with the following schema:
-   - id (INT)
-   - name (STRING)
-   - department_id (INT)
-   - employee_metadata (JSON)
+2. Upserts the data in the `employees` table.
    
 The connector expects the following secrets to be stored in the Key Vault:  
    - postgresHost
@@ -86,6 +82,20 @@ The connector expects the following secrets to be stored in the Key Vault:
    - postgresUser
    - postgresPassword
 
+## **Tables Created**
+The connector creates the following table in the database:
+- `employees` - This table is used to store employee data. The schema includes:
+  - id (INT)
+  - name (STRING)
+  - department_id (INT)
+  - employee_metadata (JSON)
+
+The table in the destination looks like this:
+
+| id | name      | department_id | employee_metadata    | _fivetran_deleted | _fivetran_synced              |
+|----|-----------|---------------|----------------------|-------------------|-------------------------------|
+| 1  | John Doe  | 1             | {"role": "Engineer"} | false             | 2025-05-11 19:53:33.959 +0000 |
+| 2  | Berry Doe | 1             | {"role": "Engineer"} | false             | 2025-05-11 19:53:33.984 +0000 |
 
 ## **Additional Considerations**
 
--- a/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
+++ b/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
@@ -2,7 +2,7 @@
 
 **Connector Overview**
 
-This connector demonstrates how to use Azure Key Vault to securely manage database credentials. It retrieves credentials from Azure Key Vault, connects to a database, and stores information about the connection in a table. 
+This connector demonstrates how to use Azure Key Vault to securely manage database credentials. It retrieves credentials from Azure Key Vault, connects to a database, fetches data from database and upserts the data. 
 
 ## **Requirements**
 
@@ -71,13 +71,13 @@ This connector uses service principal authentication to access Azure Key Vault. 
 
 The connector retrieves credentials from Azure Key Vault and uses them to establish a database connection. It then:  
 
-1. Executes a simple query to verify connectivity and retrieve database version.
-   2. Stores connection information in the "database_info" table with the following schema:
+1. Executes a simple query to verify connectivity and retrieve data from a table in the database.
+
+2. Upserts the data in the `employees` table with the following schema:
    - id (INT)
-   - host (STRING)
-   - database (STRING)
-   - connected_at (STRING)
-   - db_version (STRING)
+   - name (STRING)
+   - department_id (INT)
+   - employee_metadata (JSON)
    
 The connector expects the following secrets to be stored in the Key Vault:  
    - postgresHost
--- a/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
+++ b/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
@@ -1,0 +1,92 @@
+# **Azure Key Vault for Secret Managemen**
+
+**Connector Overview**
+
+This connector demonstrates how to use Azure Key Vault to securely manage database credentials. It retrieves credentials from Azure Key Vault, connects to a database, and stores information about the connection in a table. 
+
+## **Requirements**
+
+* [Supported Python versions](https://github.com/fivetran/fivetran_connector_sdk/blob/main/README.md#requirements)   
+* Operating System:  
+  * Windows 10 or later  
+  * macOS 13 (Ventura) or later
+
+## **Getting Started**
+
+Refer to the [Setup Guide](https://fivetran.com/docs/connectors/connector-sdk/setup-guide) to get started.
+
+## **Features**
+
+- Securely retrieves database credentials from Azure Key Vault
+- Connects to the PostgreSQL database using retrieved credentials
+- Verifies the connection by executing a simple query
+- Upserts a record into a table to confirm successful connection
+
+## **Configuration File**
+
+The connector requires the following configuration parameters:
+
+```
+{
+  "tenant_id": "<YOUR_AZURE_TENANT_ID>",
+  "client_id": "<YOUR_APP_CLIENT_ID>",
+  "client_secret": "<YOUR_APP_CLIENT_SECRET>",
+  "vault_url": "<https://YOUR-VAULT-NAME.vault.azure.net/>"
+}
+```
+
+Note: Ensure that the `configuration.json` file is not checked into version control to protect sensitive information.
+
+## **Requirements File**
+
+This connector requires Azure SDK packages to access Key Vault and connect to PostgreSQL:
+
+```
+psycopg2-binary
+azure-identity
+azure-keyvault
+azure-storage==0.36.0
+```
+
+Note: The `fivetran_connector_sdk:latest` and `requests:latest` packages are pre-installed in the Fivetran environment. To avoid dependency conflicts, do not declare them in your `requirements.txt`.
+
+## **Authentication**
+
+This connector uses service principal authentication to access Azure Key Vault. To create the necessary credentials:  
+1. Register an application in Azure Active Directory  
+   - Go to Azure Portal → Azure Active Directory → App registrations
+   - Create a new registration with a name of your choice
+   - Note the Application (client) ID and Directory (tenant) ID
+2. Create a client secret  
+   - In your app registration, go to Certificates & secrets
+   - Create a new client secret and note its value immediately
+3. Grant the application access to your Key Vault  
+   - For role-based access control: Go to Key Vault → Access control (IAM) → Add role assignment
+   - Assign the "Key Vault Secrets User" role to your application
+   - For access policy: Go to Key Vault → Access policies → Add access policy
+   - Grant "Get" and "List" permissions for secrets to your application
+
+
+## **Data Handling**
+
+The connector retrieves credentials from Azure Key Vault and uses them to establish a database connection. It then:  
+
+1. Executes a simple query to verify connectivity and retrieve database version.
+   2. Stores connection information in the "database_info" table with the following schema:
+   - id (INT)
+   - host (STRING)
+   - database (STRING)
+   - connected_at (STRING)
+   - db_version (STRING)
+   
+The connector expects the following secrets to be stored in the Key Vault:  
+   - postgresHost
+   - postgresPort
+   - postgresDatabase
+   - postgresUser
+   - postgresPassword
+
+
+## **Additional Considerations**
+
+The examples provided are intended to help you effectively use Fivetran's Connector SDK. While we've tested the code, Fivetran cannot be held responsible for any unexpected or negative consequences that may arise from using these examples. For inquiries, please reach out to our Support team.
--- a/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
+++ b/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
@@ -1,4 +1,4 @@
-# **Azure Key Vault for Secret Managemen**
+# **Azure Key Vault for Secret Management**
 
 **Connector Overview**
 
--- a/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
+++ b/examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
@@ -15,14 +15,17 @@ from azure.keyvault.secrets import SecretClient
 # Import other required libraries
 import json
 import psycopg2
-from datetime import datetime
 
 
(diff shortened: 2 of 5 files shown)


examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+# Import other required libraries
+import json
+import psycopg2
+from datetime import datetime
+
+
+# This function fetches secrets from Azure Key Vault using the Azure SDK
+# It uses the ClientSecretCredential to authenticate and SecretClient to retrieve secrets
+# The function takes a configuration dictionary as input, which contains the necessary credentials and vault URL
+# It returns a dictionary with the secrets needed to connect to the database
```
alsnkline: Can we start moving multi line comments for methods into the method
so that when you use the example the comment collapse when you collapse the method also

I think using doc strings is more pythonic.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+        except Exception as e:
+            raise RuntimeError(f"Failed to retrieve secret {secret_name}: {str(e)}")
+
+    log.info("Successfully retrieved all secrets from Azure Key Vault")
+    return ready_configuration
+
+
+# This function connects to the PostgreSQL database using the credentials fetched from Azure Key Vault
+# It takes a dictionary with the database configuration as input
+# It returns a connection object to interact with the database
```
alsnkline: move comment into method

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Define the schema function which lets you configure the schema your connector delivers.
+# See the technical reference documentation for more details on the schema function:
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#schema
+# The schema function takes one parameter:
+# - configuration: a dictionary that holds the configuration settings for the connector.
+def schema(configuration: dict):
+    # Check if the required fields are present in the configuration
+    # These fields are necessary for connecting to the Azure Key Vault
```
alsnkline, line 116: I love having this check will help running the example and debugging if you doing get it to work first time.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+    required_fields = ["tenant_id", "client_id", "client_secret", "vault_url"]
+    for field in required_fields:
+        if field not in configuration:
+            raise ValueError(f"Missing required configuration: {field}")
+
+    return [
+        {
+            "table": "database_info",
```
alsnkline: This feels confusing.
In this example the DB information is a part of getting the example to work.
The data being past is independent of the connection information or where the 'secrets are stored'

would it be clearer to have data sync'ed from postgres that is clearly nothing to do with the configuration to get to that data?

Something like customer data or orders might be more clearly totally separate.
we use those 'topic' in some of our other examples - we can likely use the same example data here also.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+        try:
+            secret = secret_client.get_secret(secret_name)
+            # Map secret names to configuration keys
+            key = secret_name.replace('postgres', '').lower()
+            ready_configuration[key] = secret.value
+        except Exception as e:
+            raise RuntimeError(f"Failed to retrieve secret {secret_name}: {str(e)}")
+
+    log.info("Successfully retrieved all secrets from Azure Key Vault")
+    return ready_configuration
```
alsnkline: This should be called db_config as that is how you are using it right?
```suggestion
    return db_config
```

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+    db_config = fetch_secrets_from_vault(configuration)
+
+    # Connect to database
+    conn = connect_to_database(db_config)
+
+    # Get database version info
+    # This is a simple query to ensure that the connection is working properly
+    with conn.cursor() as cursor:
+        cursor.execute("SELECT version()")
+        version = cursor.fetchone()[0]
```
alsnkline: Should you log.FINE some thing here (or in a few lines time) so the consumer has feedback that this check worked?
or are you thinking that the example continuing to run is enough?

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Close the connection to database
+    conn.close()
+
+    # Create data to upsert
+    # This data includes the database host, database name, connection time, and version
+    # This is a sample data structure, and you can modify it as per your requirements
+    data = {
```
alsnkline: See above - I think we should use customer or order data => something completely different to connecting to a DB or getting secrets - for clarity.

alsnkline (commented):
Love it.
I did have a couple of suggestions.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
fivetran-sahilkhirwal: I will replace all the multi line comments with docstrings for better clarity

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
fivetran-sahilkhirwal: Changed this.
Now the example does the following:
- connects to the database and fetches data from a table
- The data is employee data. It prepares the data as a dictionary
- Upserts the data to the employee table in the destination according to the schema definition

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
fivetran-sahilkhirwal: added a log to show that data has been successfully fetched from the table.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,173 @@
...
+        return conn
+    except Exception as e:
+        raise RuntimeError(f"Failed to connect to database: {str(e)}")
+
+
+# Define the schema function which lets you configure the schema your connector delivers.
+# See the technical reference documentation for more details on the schema function:
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#schema
+# The schema function takes one parameter:
+# - configuration: a dictionary that holds the configuration settings for the connector.
```
alsnkline: Should these comments be a doc string so they collapse easily when working with the example

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,173 @@
...
+    ]
+
+
+# Define the update function, which is a required function, and is called by Fivetran during each sync.
+# See the technical reference documentation for more details on the update function:
+# https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update
+# The function takes two parameters:
+# - configuration: dictionary containing any secrets or payloads you configure when deploying the connector.
+# - state: a dictionary containing the state checkpointed during the prior sync.
+#   The state dictionary is empty for the first sync or for any full re-sync.
```
alsnkline: should this be a docs string - so it collapses easily.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,173 @@
...
+    yield op.checkpoint(state)
+
+
+# This creates the connector object that will use the update function defined in this connector.py file.
+connector = Connector(update=update, schema=schema)
+
+# Check if the script is being run as the main module.
+# This is Python's standard entry method allowing your script to be run directly from the command line or IDE 'run' button.
+# This is useful for debugging while you write your code. Note this method is not called by Fivetran when executing your connector in production.
+# Please test using the Fivetran debug command prior to finalizing and deploying your connector.
```
alsnkline: should these comments be doc strings - so they are easy to close out of the way when needed.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,173 @@
...
+    sql_query = "SELECT * from sample_table"
+    with conn.cursor() as cursor:
+        cursor.execute(sql_query)
+        records = cursor.fetchall()
+        log.info(f"Fetched {len(records)} records from the database")
+
+    # Close the connection to database
+    conn.close()
+
+    for record in records:
```
alsnkline, line 161: I'm not seeing where the records are actually fetched?
I was expecting to see a method - that generated dummy data for the purposes of the example.

alsnkline (commented):
couple more thoughts

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
fivetran-sahilkhirwal:
```python
sql_query = "SELECT * from employee_table"
with conn.cursor() as cursor:
        cursor.execute(sql_query)
        records = cursor.fetchall()
        # The cursor.description attribute contains metadata about the columns in the result set.
        columns = [col[0].lower() for col in cursor.description]
        log.info(f"Fetched {len(records)} records from the database")
```
Here is where we query the database to fetch the employee data and then later we upsert this data.

Also, Added a method `insert_dummy_data_to_database` to create a dummy table and insert dummy data in it. The query then fetches the data from this dummy table and upserts it.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+        log.info("Successfully connected to database")
+        return conn
+    except Exception as e:
+        raise RuntimeError(f"Failed to connect to database: {str(e)}")
+
+
+def insert_dummy_data_to_database(conn):
+    """
+    This function inserts dummy data into the PostgreSQL database to test the connector
```
alsnkline:
```suggestion
    This function inserts dummy data into the PostgreSQL database to provide data to pull later
```

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,210 @@
...
+    This method is used only for testing purposes.
+    In a production connector, you would typically not insert dummy data into the database.
```
alsnkline:
```suggestion
    In a production connector, you would typically not need to insert data into your source the connector you could just read it.
```

I'm actually a bit uncomfortable with us including this in an example as writing directly to a source is an abusive use case - the customer would avoid paying us MAR.

I was actually thinking of a more simple method that just returned some dummy data when called.
so although we connect to the database we don't actually read any data from it....
@ameetfivetran and @fivetran-sadiemartin what do you thing about how close to working we should make this example?

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # Connect to database
+    conn = connect_to_database(db_config)
+
+    # Create dummy table and insert dummy data
+    # This method is used only for testing purposes.
+    insert_dummy_data_to_database(conn)
+
+    # Get data from the dummy table created above
+    # This is a simple query to ensure that the connection is working properly
```
alsnkline:
```suggestion
    # This is a simple query to demonstrate purpose you would usually extract all the data you want to sync
```

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
alsnkline: I'm following now.
I wasn't expecting only 1 row of data! - so I was looking for a loop around the op.upsert().

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
```diff
@@ -0,0 +1,92 @@
...
+   - department_id (INT)
+   - employee_metadata (JSON)
+   
+The connector expects the following secrets to be stored in the Key Vault:  
+   - postgresHost
+   - postgresPort
+   - postgresDatabase
+   - postgresUser
+   - postgresPassword
+
```
alsnkline, line 84: I think somewhere in the read me we would expect a representation of the data that the example would deliver to the destination.
I this case a 1 row employees table.

@fivetran-elijahdavis can share the updated template - we just added a comment to make this intent clearer.

alsnkline (commented):
few more comments.
getting closer - thank you

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,210 @@
...
+    sql_query = "SELECT * from employee_table"
+    with conn.cursor() as cursor:
```
fivetran-pranavtotala: Please create a separate method and `yield from get_data_from_database(conn)`
It will improve readability.

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
```diff
@@ -0,0 +1,210 @@
...
+    secret_names = ["postgresDatabase", "postgresHost", "postgresPassword",
+                    "postgresPort", "postgresUser"]
+
+    # Fetch the secrets
+    db_config = {}
+    for secret_name in secret_names:
+        try:
+            secret = secret_client.get_secret(secret_name)
+            # Map secret names to configuration keys
+            key = secret_name.replace('postgres', '').lower()
```
varundhall: this looks like complex logic

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/README.md
fivetran-elijahdavis: After data hanlding ~ 

## **Tables Created**

*Summary of Tables replicated.*

*Screenshot of table showing sample data. (Where applicable...no PII.)*

Full structure here - https://github.com/fivetran/fivetran_connector_sdk/pull/135/files

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
fivetran-sahilkhirwal: Simplified the logic

varundhall (approved):

fivetran-pranavtotala (approved):
LGTM

examples/common_patterns_for_connectors/azure_keyvault_for_secret_management/connector.py
ameetfivetran: This example is to demonstrate azure keyvault usecase. Reading/writing into source is for demonstrating successful connection establishment. For someone to test this, they may use the dummy data insertion aswell. but we can comment it out if you think customers will directly use the examples in some cases.

alsnkline: Yeah I get why its here and it does enable us to validate that the example works end to end.
However I think the risk of including code that shows how to avoid paying MAR for the pipeline is too risky to include in our examples repo.

I think we need to remove the code that writes the dummy data to the DB.

Instead we should do 1 of three things:
* fake/move the database the the playground pypi - so when you call it the playground responds with the data
* comment out the database calling code and add a call to a method to get dummy data as if the database read had returned.
* use Azure Keyvault to get credentials for an API rather than a Database

fivetran-pranavtotala (approved):
LGTM

PR #137, opened 2025-05-06
Feature(Connector-sdk): Adding example for Neo4j graph database

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-945668

### What the example does

- Fetches the data from neo4j graph database
- upserts the data to destination

### Validation
Screenshot 2025-05-07 at 03 07 37

Screenshot 2025-05-07 at 03 10 32


examples/source_examples/neo4j/connector.py
```diff
@@ -0,0 +1,220 @@
...
+
+    results = session.execute_read(
+        lambda tx: tx.run(cypher_query).data()
+    )
+
+    # Process each user record
+    for record in results:
+        # Convert any datetime objects to ISO format strings
+        for key, value in record.items():
+            if isinstance(value, (datetime.datetime, datetime.date)):
```
fivetran-bhargavpatel: In the users table, there is no column present with type datetime. Why this conversion required?

fivetran-sahilkhirwal: This is to show if there are columns with datetime, the customers can handle it in this method. Any data processing logic can be implemented here based on the needs

fivetran-satvikpatil (approved):

PR #138, opened 2025-05-07
Feature(connector-sdk): Adding clickhouse database example

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-945673

### What the example does
- connects to clickhouse database and fetches the data in streaming fashion for better memory management
- Upserts the record as soon as record is fetched

### Validation

Screenshot 2025-05-07 at 14 09 49
Screenshot 2025-05-07 at 14 09 58
Screenshot 2025-05-07 at 14 10 10


examples/source_examples/clickhouse/connector.py
```diff
@@ -0,0 +1,147 @@
...
+
+def create_clickhouse_client(configuration: dict):
+    """
+    Create and test a connection to ClickHouse
+    This function creates a ClickHouse client using the clickhouse_connect library.
+    Args:
+        configuration: a dictionary that holds the configuration settings for the connector.
+    Returns:
+        client: a ClickHouse client object
+    """
```
fivetran-bhargavpatel, line 29: # check if required configuration values are present in configuration
    for key in ['username', 'password', 'host']:
        if key not in configuration:
            raise ValueError(f"Missing required configuration value : {key}")
            
Can we add this kind validation here? just like added in Neo4j example.

fivetran-sahilkhirwal:
```python
    # check if required configuration values are present in configuration
    for key in ['hostname', 'username', 'password', 'database']:
        if key not in configuration:
            raise ValueError(f"Missing required configuration value : {key}")
```
Added this in the `schema()` method which ensures that it runs before anything else

fivetran-satvikpatil (approved):

PR #139, opened 2025-05-08
Feature(Connector-sdk) : Example for Cassandra database

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-945664

### What the example does : 
- Connects to cassandra 
- Inserts sample data after creating sample keyspace and table
- Fetches the data in a paginated manner to ensure data is not loaded completely in the memory
- Upserts the data

### Validation
Screenshot 2025-05-08 at 20 11 36

### DIFF
--- a/examples/source_examples/cassandra/connector.py
+++ b/examples/source_examples/cassandra/connector.py
@@ -1,0 +1,271 @@
+# This is an example for how to work with the fivetran_connector_sdk module.
+# It defines a simple 'update' method, which upserts data from Cassandra database.
+# See the Technical Reference documentation (https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update)
+# and the Best Practices documentation (https://fivetran.com/docs/connectors/connector-sdk/best-practices) for details
+
+# Import the required classes from the connector SDK
+from fivetran_connector_sdk import Connector
+from fivetran_connector_sdk import Operations as op
+from fivetran_connector_sdk import Logging as log
+
+# Import the required libraries
+import json
+import uuid
+from datetime import datetime, timezone
+from dateutil import parser
+# Import the Cassandra drivers
+from cassandra.cluster import Cluster
+from cassandra.auth import PlainTextAuthProvider
+
+
+def create_cassandra_session(configuration: dict):
+    """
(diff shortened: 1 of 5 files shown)


examples/source_examples/cassandra/connector.py
```diff
@@ -0,0 +1,271 @@
...
+        # Create a Cassandra session with the provided configuration
+        cluster = Cluster([host], port=port, auth_provider=auth_provider)
+        session = cluster.connect()
+        log.info("Connection to Cassandra successful")
+        return session
+    except Exception:
+        raise RuntimeError("Failed to connect to Cassandra")
+
+
+def create_dummy_keyspace_and_table(session, keyspace, table_name):
```
manjutapali: let's move adding dummy data method to another python file.

examples/source_examples/cassandra/connector.py
```diff
@@ -0,0 +1,271 @@
...
+            id UUID PRIMARY KEY,
+            name text,
+            created_at timestamp
+            )"""
+        )
+
+    log.info("Created table successfully")
+
+
+def insert_dummy_data(session, keyspace, table_name, record_count=10):
```
manjutapali: let's move adding dummy data method to another python file.

manjutapali (changes requested):
added minor comment

examples/source_examples/cassandra/connector.py
```diff
@@ -0,0 +1,207 @@
+# This is an example for how to work with the fivetran_connector_sdk module.
+# It defines a simple 'update' method, which upserts data from Cassandra database.
```
fivetran-satvikpatil: This also defines schema method. Instead of this point mention just what does this example do.
Something like: 
`This example demonstrates how to work with a Cassandra database containing a large dataset and using pagination.`
You can improve this.

manjutapali (approved):
Looks good.

Should we get review from tech writer?

fivetran-satvikpatil (approved):

examples/source_examples/cassandra/connector.py
fivetran-sahilkhirwal: Updated this

PR #140, opened 2025-05-12
Feature(Connector-sdk): Adding microsoft excel file example

fivetran-sahilkhirwal:
closes https://fivetran.height.app/T-929197

### What example does:
- Fetches the excel file from s3 bucket
- upserts the data from excel using - `pandas`, `python-calamine` and `openpyxl` libraries

### Validation
Screenshot 2025-05-13 at 02 58 21
Screenshot 2025-05-13 at 02 58 39
Screenshot 2025-05-13 at 02 58 59
Screenshot 2025-05-13 at 03 03 18
Screenshot 2025-05-13 at 03 03 35
Screenshot 2025-05-13 at 03 03 51

### DIFF
--- a/examples/source_examples/microsoft_excel/README.md
+++ b/examples/source_examples/microsoft_excel/README.md
@@ -1,0 +1,105 @@
+# Microsoft Excel File Example
+
+## Connector overview
+
+This connector fetches Microsoft Excel files from AWS S3 and processes them using three different methods to extract the data:
+- `pandas` library,
+- `python-calamine` library, and
+- `openpyxl` library.
+
+It's designed to efficiently handle Excel files of various sizes, providing flexibility for different data processing needs.
+
+## Requirements
+
+* [Supported Python versions](https://github.com/fivetran/fivetran_connector_sdk/blob/main/README.md#requirements)   
+* Operating System:  
+  * Windows 10 or later  
+  * macOS 13 (Ventura) or later
+
+## Getting started
+
+Refer to the [Setup Guide](https://fivetran.com/docs/connectors/connector-sdk/setup-guide) to get started.
+
+## Features
+
+- Retrieves Excel file from AWS S3 buckets
+- Processes Excel data using three different methods:
+  - Pandas - suitable for small to medium-sized files
+  - Python-calamine - memory-efficient for larger files
+  - Openpyxl - read-only mode for memory efficiency
+- Creates three destination tables with identical schemas
+- Handles proper cleanup of temporary files
+
+## Configuration file
+
+The connector requires AWS S3 credentials and file information:
+
+```
+{
+  "aws_access_key_id": "<YOUR_AWS_ACCESS_KEY>",
+  "aws_secret_access_key": "<YOUR_AWS_SECRET_KEY>",
+  "region_name": "<YOUR_AWS_REGION>",
+  "bucket_name": "<YOUR_S3_BUCKET>",
+  "file_name": "<path/to/your/file.xlsx>"
+}
+```
+
+Note: Ensure that the `configuration.json` file is not checked into version control to protect sensitive information.
+
+## Requirements file
+
+The connector requires the following Python packages:
+
+```
+boto3==1.38.14
+pandas==2.2.3
+openpyxl
+python-calamine
+```
+
+In order to use the `calamine` engine with `pandas` library, you need to include the `python-calamine` package and ensure that the `pandas` version is `equal to or above 2.2`.
+
+Note: The `fivetran_connector_sdk:latest` and `requests:latest` packages are pre-installed in the Fivetran environment. To avoid dependency conflicts, do not declare them in your `requirements.txt`.
+
+## Authentication
+
+This connector authenticates with AWS using an access key ID and secret access key. These credentials must have permissions to read objects from the specified S3 bucket. You can create these credentials in the AWS IAM console.  To set up authentication:  
+- Create an IAM user with S3 read access
+- Generate an access key and secret key for this user
+- Add these credentials to your configuration.json file
+
+## Data handling
+
+The connector processes Excel files using three different methods:  
+1. Pandas (Refer to `upsert_using_pandas` function): Loads the entire Excel file into memory, which is suitable for small to medium-sized files.  
+2. Python-calamine (Refer to `upsert_using_calamine` function): Uses the calamine engine which is more memory-efficient and recommended for large files.  
+3. Openpyxl (Refer to `upsert_using_openpyxl` function): Uses read-only mode for better memory efficiency when processing large files.  
+
+Data from each method is upserted into separate tables with identical schemas.
+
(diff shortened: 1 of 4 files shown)


examples/source_examples/microsoft_excel/connector.py
```diff
@@ -0,0 +1,284 @@
...
+    for row in rows:
+        # Convert the row to a dictionary
+        # Each row is a dictionary with column names as keys
+        record = dict(zip(headers, row))
+        # Upsert the record into the destination table
+        yield op.upsert(table="excel_data_openpyxl", data=record)
+
+    # Save the progress by checkpointing the state.
+    yield op.checkpoint(state)
+    wb.close()
```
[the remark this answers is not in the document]
fivetran-sahilkhirwal: Screenshot 2025-05-13 at 03 13 06
Throws error when trying to do this way

examples/source_examples/microsoft_excel/connector.py
```diff
@@ -0,0 +1,284 @@
...
+    This is suitable for small to medium-sized files as it loads the entire file into memory.
+    This is not recommended for large files, because it can cause Out of Memory errors.
+    Args:
+        temp_filename: The path to the temporary file containing the Excel data.
+        state:  a dictionary containing the state checkpointed during the prior sync.
+    """
+
+    # Read Excel file using pandas from the temporary file
+    # This loads the entire Excel file into a pandas DataFrame
+    # This is not recommended for large files, because it can cause Out of Memory errors
```
fivetran-pranavtotala, line 93: good point!

examples/source_examples/microsoft_excel/connector.py
```diff
@@ -0,0 +1,284 @@
...
+
+    # Download the Excel file from S3 to a temporary file
+    temp_filename = download_excel_file(s3_client, bucket_name, file_key)
+
+    try:
+        # Process the Excel file using pandas and upsert the data
+        yield from upsert_using_pandas(temp_filename, state)
+
+        # Process the Excel file using python-calamine and upsert the data
+        yield from upsert_using_calamine(temp_filename, state)
```
fivetran-pranavtotala, line 274: We are using same file for the upsert, the example should mention at the start off the try catch
- that any of the upsert functions will work.
- which is the fastest and the recommended method. (should be present here too, as it will be easier to decide if the comment is near the choice)

examples/source_examples/microsoft_excel/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Add these credentials to your configuration.json file
+
+## Data handling
+
+The connector processes Excel files using three different methods:  
+1. Pandas (Refer to `upsert_using_pandas` function): Loads the entire Excel file into memory, which is suitable for small to medium-sized files.  
+2. Python-calamine (Refer to `upsert_using_calamine` function): Uses the calamine engine which is more memory-efficient and recommended for large files.  
+3. Openpyxl (Refer to `upsert_using_openpyxl` function): Uses read-only mode for better memory efficiency when processing large files.  
+
+Data from each method is upserted into separate tables with identical schemas.
```
fivetran-pranavtotala:
```suggestion
Data from each method is identical, but upserted into separate tables with identical schemas.
```

fivetran-pranavtotala:
```suggestion
Data from each method is identical, but upserted into separate tables with identical schemas.
```

fivetran-pranavtotala (approved):
LGTM

examples/source_examples/microsoft_excel/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Refer to the [Setup Guide](https://fivetran.com/docs/connectors/connector-sdk/setup-guide) to get started.
+
+## Features
+
+- Retrieves Excel file from AWS S3 buckets
+- Processes Excel data using three different methods:
+  - Pandas - suitable for small to medium-sized files
+  - Python-calamine - memory-efficient for larger files
+  - Openpyxl - read-only mode for memory efficiency
```
manjutapali: The python-calamine is faster in processing the file. Add details on fast processing.

examples/source_examples/microsoft_excel/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Getting started
+
+Refer to the [Setup Guide](https://fivetran.com/docs/connectors/connector-sdk/setup-guide) to get started.
+
+## Features
+
+- Retrieves Excel file from AWS S3 buckets
+- Processes Excel data using three different methods:
+  - Pandas - suitable for small to medium-sized files
+  - Python-calamine - memory-efficient for larger files
```
manjutapali: I don't think calamine package is memory efficient - https://hakibenita.com/fast-excel-python.

Screenshot 2025-05-20 at 1 05 12 PM

manjutapali: verify for Openpyxl too.

manjutapali (changes requested):
added comments

examples/source_examples/microsoft_excel/README.md
fivetran-sahilkhirwal: Based on the documentations of `openpyxl` and `python-calamine`:
- `python-calamine` does load the entire data into the memory before parsing as the rust binding does not support the streaming. but the memory overhead of `python-calamine` is very low. The peak memory usage of `python-calamine` is around the size of the raw data. It should be able to easily handle files with size 500MB or less.
- `openpyxl` has read only mode which supports streaming. It does not build the entire workbook and only keeps one row in the memory at a time. This reduces the memory utilisation. But `openpyxl` adds overheads due to absence of in-memory caching.

So, To summarize, `python-calamine` loads the entire workbook before iterating over it but has negligible overhead and processes data faster due to this. Whereas, `openpyxl` does not load the entire workbook and keeps only one row in memory at a time. But `openpyxl` also has some overhead due to it and performs slower than `python-calamine`.

> One major difference between the libraries is that openpyxl’s `read-only` mode opens a workbook almost immediately making it suitable for multiple processes, this also reduces memory use significantly. `xlrd` does also not automatically convert dates and times into Python datetimes, though it does annotate cells accordingly but to do this in client code significantly reduces performance.

Thus, `python-calamine` takes more memory, has less over head and faster processing than `pandas` or `openpyxl`.
`Openpyxl` takes least memory, has more overhead and is slower than `python-calamine`

fivetran-sahilkhirwal: Based on the documentations, made changes in the README to highlight the processing speed and memory consumption of each method.

manjutapali: We can update the doc, recommending not to use calamine incase of >1GB files. As this is still going to cause OOM issues.

fivetran-sahilkhirwal: Added this in the doc and comments

PR #176, opened 2025-06-17
Update README.md

fivetran-dejantucakov:
We need a full URL so as not to have a 404 in the embedded page on Fivetran docs.


5tran-alexil (approved):
LGTM

PR #232, opened 2025-08-11
fix(codeowners): update with specific file types for Team Menon

varundhall:
### Jira ticket
Link https://fivetran.atlassian.net/browse/RD-968977

### Description of Change
Update Codeowners with certain file types instead of too broad coverage.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,9 @@
-* @fivetran/menon
+**/*.py @fivetran/menon
+**/*.yml @fivetran/menon
+**/*.yaml @fivetran/menon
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,2 +1,9 @@
-* @fivetran/menon
+**/*.py @fivetran/menon
```
fivetran-satvikpatil, line 1: Why not readme changes? Shouldn't menon own everything in repo?

varundhall: Readme should be handled by Tech Writers, they anyways loop Menon if something needs clarity.

PR #275, opened 2025-09-16
fix(workflows): Add Github API permissions for fork PRs

fivertran-karunveluru:
### Description of Change

Fixes `403 Error`

```
    data: {
      message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/issues/comments#create-an-issue-comment',
      status: '403'
    }
```

### Checklist

- ~[] Tested the connector with `fivetran debug` command.~
- ~[ ] Added/Updated example specific README.md file, [refer here](https://github.com/fivetran/fivetran_connector_sdk/tree/main/template_example_connector/README_template.md) for template.~
- ~[ ] Followed Python Coding Standards, [refer here](https://fivetran.slab.com/posts/connector-sdk-examples-pr-policy-and-python-coding-standards-yzr9ggss)~

### DIFF
--- a/.github/workflows/code-quality-check.yml
+++ b/.github/workflows/code-quality-check.yml
@@ -3,6 +3,12 @@ name: Python
 on:
   pull_request:
 
+permissions:
+  contents: read
+  pull-requests: read
+  issues: write
+  actions: read
+
 jobs:
   code-quality-check:
     name: Check Code Quality


fivertran-karunveluru:
@varundhall @fivetran-satvikpatil Let me know if you need my assistance testing this

fivertran-karunveluru:
> @fivetran-satvikpatil I don't think this will work due to our token in the CI action, we should check it once. Maybe stop posting comment if its a fork.

I think this approach might be addressing the symptom rather than the root cause. Adding fork-specific logic would increase the maintenance burden and complexity without fully addressing the underlying token permissions issue.

fivetran-sahilkhirwal:
Hi @fivertran-karunveluru 
This approach should not work as the secrets in CI are not exposed for `pull_request` type of workflow triggers to prevent security risks from any malicious code in the fork. This is the default behaviour of github.
All the github workflows work correctly when the PR is raised from the main repo instead of fork. Adding the permissions will not expose the secrets which are causing the workflows to fail for PRs from the fork. You can close this PR. 
One potential fix for this is to trigger it on `pull_request_target` which exposes the secrets to the workflow. The fix for this is [underway](https://github.com/fivetran/fivetran_connector_sdk/pull/371)

PR #478, opened 2025-11-26
Update ERD image link in README.md

5tran-alexil:
### Jira ticket
Closes https://fivetran.atlassian.net/browse/RD-1055647

### Description of Change
ERD link updated to absolute path since it's embedded in Fivetran live docs


fivetran-dejantucakov (approved):
@5tran-alexil Tnx for looking into this!

PR #483, opened 2025-11-30
feature(sdk): Clarify configuration loading behavior in quickstart ex…

fivetran-ameysharma:
Closes https://fivetran.atlassian.net/browse/RD-969244

This PR improves the __main__ block comment in the quickstart example connector to better explain how configuration is loaded when running the script directly.

**Key updates:**

- Added a clear explanation that the example loads configuration.json to work smoothly with IDE Run/Debug.
- Added a pointer to the documentation for advanced users who want to pass custom configuration paths during IDE-based debugging.
- Preserved the original behavior so the example continues to run out-of-the-box without additional setup.

This provides better clarity for developers using the example for local debugging, without adding complexity to the example itself.

### DIFF
--- a/examples/quickstart_examples/configuration/connector.py
+++ b/examples/quickstart_examples/configuration/connector.py
@@ -97,15 +97,19 @@ def update(configuration: dict, state: dict):
 # This creates the connector object that will use the update and schema functions defined in this connector.py file.
 connector = Connector(update=update, schema=schema)
 
-# Check if the script is being run as the main module.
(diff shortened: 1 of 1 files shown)


fivetran-ameysharma:
> @fivetran-ameysharma why are we making change only in one example, please check with @alsnkline, if she is aligned with your changes, we should batch update all the examples in one PR.

@varundhall we have discussed. We do not want to make changes in all. This example link is present in our Main Doc

Screenshot 2025-12-01 at 4 24 00 PM


And we have Main PR as well https://github.com/fivetran/engineering/pull/254873

PR #489, opened 2025-12-15
feature(connector_sdk): Updated sub agent files

fivetran-rishabhghosh:
### Jira ticket
Closes https://fivetran.atlassian.net/browse/RD-1069730

### Description of Change
- Update Claude sub agent files
- Update agent md files for cursor, vscode
- Delete windsurf

### Testing
NA

### Checklist

- [ ] Tested the connector with `fivetran debug` command.
- [ ] Added/Updated example-specific README.md file, see [the README template](https://github.com/fivetran/fivetran_connector_sdk/tree/main/template_connector/README_template.md) for the required structure and guidelines.
- [ ] Followed Python Coding Standards, [refer here](https://github.com/fivetran/fivetran_connector_sdk/blob/main/PYTHON_CODING_STANDARDS.md)

### DIFF
--- a/all_things_ai/ai_agents/AGENTS.md
+++ b/all_things_ai/ai_agents/AGENTS.md
@@ -18,11 +18,11 @@ You are a specialized AI assistant focused on helping users build, test, and val
 - Authentication and security protocols
 - AI/ML data pipeline patterns
 - Reference Documentation:
-  * [Fivetran Connector SDK Documentation](https://fivetran.com/docs/connector-sdk)
-  * [SDK Examples Repository](https://github.com/fivetran/fivetran_connector_sdk/tree/main/examples)
-  * [Technical Reference](https://fivetran.com/docs/connector-sdk/technical-reference)
-  * [Supported Datatypes](https://fivetran.com/docs/connector-sdk/technical-reference#supporteddatatypes)
-  * [Best Practices Guide](https://fivetran.com/docs/connector-sdk/best-practices)
+  - [Fivetran Connector SDK Documentation](https://fivetran.com/docs/connector-sdk)
+  - [SDK Examples Repository](https://github.com/fivetran/fivetran_connector_sdk/tree/main/examples)
+  - [Technical Reference](https://fivetran.com/docs/connector-sdk/technical-reference)
+  - [Supported Datatypes](https://fivetran.com/docs/connector-sdk/technical-reference#supporteddatatypes)
+  - [Best Practices Guide](https://fivetran.com/docs/connector-sdk/best-practices)
 
 ## Technical Requirements
 
--- a/all_things_ai/ai_agents/claude_code/.claude/agents/ft-csdk-ask.md
+++ b/all_things_ai/ai_agents/claude_code/.claude/agents/ft-csdk-ask.md
@@ -38,11 +38,11 @@ This agent specializes in:
 - Code analysis and pattern recognition
 - Best practices and optimization strategies
 - Reference Documentation:
-  * [Fivetran Connector SDK Documentation](https://fivetran.com/docs/connector-sdk)
-  * [SDK Examples Repository](https://github.com/fivetran/fivetran_connector_sdk/tree/main/examples)
-  * [Technical Reference](https://fivetran.com/docs/connector-sdk/technical-reference)
-  * [Supported Datatypes](https://fivetran.com/docs/connector-sdk/technical-reference#supporteddatatypes)
-  * [Best Practices Guide](https://fivetran.com/docs/connector-sdk/best-practices)
+  - [Fivetran Connector SDK Documentation](https://fivetran.com/docs/connector-sdk)
+  - [SDK Examples Repository](https://github.com/fivetran/fivetran_connector_sdk/tree/main/examples)
+  - [Technical Reference](https://fivetran.com/docs/connector-sdk/technical-reference)
+  - [Supported Datatypes](https://fivetran.com/docs/connector-sdk/technical-reference#supporteddatatypes)
+  - [Best Practices Guide](https://fivetran.com/docs/connector-sdk/best-practices)
 
 ---
 
--- a/all_things_ai/ai_agents/claude_code/.claude/agents/ft-csdk-fix.md
+++ b/all_things_ai/ai_agents/claude_code/.claude/agents/ft-csdk-fix.md
@@ -20,10 +20,10 @@ This agent specializes in:
 - Debugging and troubleshooting patterns
 - Error classification and root cause analysis
 - Reference Documentation:
-  * [Fivetran Connector SDK Documentation](https://fivetran.com/docs/connector-sdk)
-  * [SDK Examples Repository](https://github.com/fivetran/fivetran_connector_sdk/tree/main/examples)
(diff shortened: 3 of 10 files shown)


fivetran-sahilkhirwal (dismissed):

5tran-alexil:
@fivetran-rishabhghosh Can we regenerate the update provided it's done with AI, since Copilot does not picks the instructions/rules from the PR?
```
## Never use:
- Title Case in subheadings
- Nested bulleted lists where a simple list would be sufficient
- Bold text to stress important information. If you see bold text, you should flag this and request to change it to plain text unless this is a UI element name.
Example of wrong format and style:

"- **API example**
   - This example shows how to sync data using an API key and secret."

The correct format and style would be as follows:

"- API example - This example shows how to sync data using an API key and secret."

## Headings:

- There should be only one H1 heading in the file, and it should use Title Case, for example: # Connector SDK Somecompany API Connector Example
- All other headings should use Sentence case, for example: ## Requirements file
- Do not use numbers in headingds: WRONG: "## 4. Best practices", CORRECT: "## Best practices"

## Lists:

Use bulleted lists when just listing items in no particular order or when listing the alternatives, for example:
"The management actions are as follows:
- Create connection
- Retrieve connection
- Delete connection"

Use numbered list for actions that need to be taken in a particular order, for example:
"To obtain your API key, do the following:
1. Log in to your DreemBleem account.
2. Go to **Settings > API Key**.
3. Make a note the API key and secret. You will need to add it to `configuration.json` file later."
```

fivetran-rishabhghosh:
@5tran-alexil I passed the same to Claude, and it suggests not applying these changes to the agent MD files. 
Screenshot 2025-12-16 at 3 57 43 PM

Even I am aligned that these rules apply to user facing docs and these agent files although can be read by customers but are meant for AI Agents

5tran-alexil:
It turns to a battle between "AI's" opinions, here's what ChatGPT output to my question how does formatting of text in an instruction Markdown file affect the output and if LLMs actually recognize/react to formatting:

Short version: for *this kind of agent file*, title case, bold, and heading numbering matter **a lot for humans**, and **only a little** for the model. The model’s behavior is driven ~95% by the actual wording and order of the instructions, not whether section titles are `PRIMARY ROLE` vs `Primary role` or bold vs plain.

More detail, mapped to your example.

---

## 1. Heading style & Title Case

Things like:

```markdown
## Core Identity and Purpose
1. PRIMARY ROLE
2. KNOWLEDGE BASE
```

vs

```markdown
## Core identity and purpose
1. Primary role
2. Knowledge base
```

To the model, those are basically the same:

* All the important tokens – “core”, “identity”, “purpose”, “role”, “knowledge base” – are still there.
* Changing `PRIMARY` → `Primary` doesn’t change the semantics at all.
* Markdown `##` / `1.` are just extra tokens the model has seen millions of times; they don’t have a magic “this is higher priority” flag.

So: **you can safely normalize case and heading style** for this file, as long as you keep the same words and order.

---

## 2. Bold text

Example:

```markdown
2. LOGGING - CRITICAL: Use EXACT method names
- **CORRECT:** `log.info()`, ...
- **WRONG:** `log.error()` (does NOT exist in Fivetran SDK)
```

If you drop the bold:

```markdown
- CORRECT: `log.info()`, ...
- WRONG: `log.error()` (does NOT exist in Fivetran SDK)
```

The model still sees “CORRECT”, “WRONG”, and the code examples. Bold adds `**` tokens, which the model *has learned are often used to mark important bits*, but they’re a **small** signal compared to:

* Explicit words: “CRITICAL”, “ONLY”, “NEVER”, “ALWAYS”, “MUST”.
* Negative examples: “WRONG: `log.error()` (does NOT exist…)”.

So: removing bold **probably does not meaningfully change behavior**, but if you’re using bold to visually cluster *critical constraints*, it’s cheap to keep. Think of it as a tiny extra nudge plus a big boost for human readers.

---

## 3. Numbering & nested lists

Numbered sections and nesting like:

```markdown
## Technical Requirements

### Runtime Environment
- **Memory:** 1 GB RAM
- **CPU:** 0.5 vCPUs
```

vs a flatter version:

```markdown
## Technical requirements

Runtime environment:
- Memory: 1 GB RAM
- CPU: 0.5 vCPUs
```

For a single Markdown file passed as one system prompt:

* The model doesn’t actually execute “Step 1, Step 2…” — it just reads them as tokens.
* Flattening / renumbering won’t change behavior *as long as you keep*:

  * the same sentences,
  * in roughly the same order,
  * with lists still visibly lists (bullets or line breaks).

The risk with flattening/nesting comes more from **human clarity**:

* If flattening makes constraints easier to misread or merge (“oh that bullet belongs to the logging section, not schema”), humans editing later might weaken or delete something important → *that* changes the agent.

But the model itself doesn’t “parse a tree” of your Markdown headings; it’s just reading text in sequence.

---

## 4. ALL CAPS inside rules (`NEVER`, `ALWAYS`, `CRITICAL`)

This is the one formatting dimension I wouldn’t casually nerf.

The model has seen **a ton** of text like:

* “**IMPORTANT:** NEVER share passwords.”
* “CRITICAL: ALWAYS sanitize user input.”

It has learned that words like `NEVER`, `ALWAYS`, `MUST NOT`, `CRITICAL` — especially in caps — are often *hard constraints*. Lowercasing them (and especially removing them) is more likely to reduce the perceived strength of the rule than changing heading case.

So my practical advice:

* Keep **strong modals** and caps where they encode semantics:

  * `NEVER use op.Operation`
  * `ALWAYS use simple dict and list type hints`
* You can still follow your style guide by e.g. using sentence case headings but keeping caps inside the rule sentence itself.

---

## 5. So, will the agent’s output actually change?

If you:

* keep the **same words** and examples,
* in the **same order**,
* keep strong constraint phrases (`NEVER`, `ALWAYS`, `MUST`, `CRITICAL`) in the body,

then switching from “noisy” Markdown to “bland, consistent” Markdown (title case, normalized headings, less bold) will:

* **Barely affect behavior**, if at all, for this kind of agent file.
* Definitely improve maintainability for humans, which indirectly improves agent quality over time.

If you *also* watered down language:

* `NEVER use op.Operation` → `Avoid using op.Operation`
* `WRONG` → `Not recommended`

then yes, I’d expect more drift in behavior than from any change in bold/heading style.

---

### TL;DR for your repo

For `.claude/agents/*.md` like the snippet you showed, I’d feel safe to:

* Normalize heading capitalization (`PRIMARY ROLE` → `Primary role`).
* Standardize numbering style and list indentation.
* Remove non-essential bold like `**Memory:**` / `**CPU:**`.

But I’d **keep**:

* Caps + strong modals for hard rules:

  * `NEVER use op.Operation in type hints`
  * `ALWAYS use simple dict and list`
* Very clear “CORRECT” vs “WRONG” examples.

That gets you closer to style-guide compliance without sacrificing the tiny-but-useful salience signals that actually matter for the model.

5tran-alexil:
I'd say, at least emojis should go and * replaced by - in bulleted lists. Title Case and bolding and numbering in headings as well. And what do we stress for an agent in cases like **Memory**, **CPU**? There's no real **EMPHASIS IF _EVERYTHING_ IS BOLD AND UPPER CASE**. 
What is the point in ## **5. BEST PRACTICES**?

fivetran-sahilkhirwal (approved):

PR #503, opened 2026-01-15
Fix duplicate MAR issue in Toast connector for applied tax records

frederikkofoed:
**Problem**
The orders_check_selection_applied_tax table generates a random UUID for records that lack a guid from the Toast API. Since this random ID is part of the primary key, every sync treats the same record as a new unique row, causing massively inflated Monthly Active Row (MAR) counts in Fivetran.

**Solution**
Changed to use a deterministic hash based on orders_check_selection_id and taxRate_id, ensuring the same record gets the same ID across syncs.

Changes
- Added import hashlib (standard library, already used in other connectors in this repo)
- Replaced uuid.uuid4() with an MD5 hash of parent_id + tax_rate_id in process_child() for orders_check_selection_applied_tax records

### DIFF
--- a/connectors/toast/connector.py
+++ b/connectors/toast/connector.py
@@ -627,7 +627,6 @@ def make_headers(conf, base_url, state, key):
         except Exception as e:
             print(f"⚠️ Token decryption failed: {e}, re-authenticating...")
 
-
     # No valid token OR token expiring within 1 hour, request a new one
     payload = {
         "clientId": conf.get("clientId"),


connectors/toast/connector.py
```diff
@@ -564,8 +565,12 @@ def process_child(parent, table_name, id_field_name, id_field):
             # log.fine(f"flattening fields in {table_name}")
             p = flatten_fields(fields_to_flatten[table_name], p)
         # check for null guids in appliedTaxes[]
+        # Use deterministic ID based on parent selection + tax rate to ensure consistent identification across syncs
         if table_name == "orders_check_selection_applied_tax" and p.get("guid") is None:
-            p["guid"] = "gen-" + str(uuid.uuid4())
+            parent_id = p.get("orders_check_selection_id", "")
+            tax_rate_id = p.get("taxRate_id", "")
```
manjutapali: @frederikkofoed Can `parent_id` and `tax_rate_id` will be None?

frederikkofoed: Good question. Looking at the Toast API docs:

  - `orders_check_selection_id` is always present
  - `taxRate.guid` is optional and could be None

  To handle edge cases where multiple applied taxes on the same selection might lack a `taxRate_id`, we could include additional fields in the hash. Would this work?

  ```python
  if table_name == "orders_check_selection_applied_tax" and p.get("guid") is None:
      parent_id = p.get("orders_check_selection_id") or ""
      tax_rate_id = p.get("taxRate_id") or ""
      tax_name = p.get("name") or ""
      tax_rate = str(p.get("rate") or "")
      unique_string = f"{parent_id}_{tax_rate_id}_{tax_name}_{tax_rate}"
      p["guid"] = "gen-" + hashlib.md5(unique_string.encode()).hexdigest()
```

Comment (manjutapali) [reply connectors/toast/connector.py]: If the `tax_name` and `tax_rate` is updated, it will create new entry.

Comment (frederikkofoed) [reply connectors/toast/connector.py]: Fair point about name and rate changing. However, these are applied tax records on completed orders. They're immutable historical snapshots of what was charged at the time of the transaction. The name and rate on an applied tax record won't change after the order is placed, so including them in the hash should be safe.

And even in the worst case, this is still a massive improvement over the current behavior where a random UUID is generated on every sync. In our case,  a few hundred actual records were being charged as several million MARs over a few months.

I've pushed an updated commit that includes name and rate in the hash along with the `or ""` handling for None values.

**Review by fivetran-pranavtotala — APPROVED** [review]
LGTM

File: connectors/toast/connector.py [hunk 2]
```diff
@@ -13,6 +13,7 @@
 import json
 import copy
 import uuid
```
fivetran-pranavtotala: remove this, not used.

fivetran-pranavtotala (approved):
LGTM, do fix the flake8 python test.

manjutapali (dismissed):

frederikkofoed:
Fixed the pre-existing flake8 W503 warnings that were causing the CI test to fail.

connectors/toast/connector.py
fivetran-pranavtotala:
```suggestion
        if state["token_ttl"] > current_time + 3600:
            try:
                auth_token = fernet.decrypt(state["encrypted_token"].encode()).decode()
                log.info("encrypted_token found with at least an hour left, reusing")
                return {
                    "Authorization": f"Bearer {auth_token}",
                    "Accept": "application/json",
                }, state
            except Exception as e:
                print(f"⚠️ Token decryption failed: {e}, re-authenticating...")

```

connectors/toast/connector.py
```diff
@@ -613,9 +619,9 @@ def make_headers(conf, base_url, state, key):
 
     # Check if a valid token exists and is not expiring in the next hour
     if (
```
fivetran-pranavtotala:
```suggestion
    if "encrypted_token" in state and "token_ttl" in state:
```

connectors/toast/connector.py
```diff
@@ -612,17 +618,27 @@ def make_headers(conf, base_url, state, key):
     current_time = time.time()
 
     # Check if a valid token exists and is not expiring in the next hour
-    if (
-        "encrypted_token" in state
-        and "token_ttl" in state
-        and state["token_ttl"] > current_time + 3600
+    if "encrypted_token" in state and "token_ttl" in state:
```
fivetran-pranavtotala:
```suggestion
    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
```

connectors/toast/connector.py
```diff
@@ -612,17 +618,27 @@ def make_headers(conf, base_url, state, key):
...
+            try:
+                auth_token = fernet.decrypt(state["encrypted_token"].encode()).decode()
+                log.info("encrypted_token found with at least an hour left, reusing")
+                return {
+                    "Authorization": f"Bearer {auth_token}",
+                    "Accept": "application/json",
+                }, state
+            except Exception as e:
+                print(f"⚠️ Token decryption failed: {e}, re-authenticating...")
+
```
fivetran-pranavtotala:
```suggestion
            print(f"⚠️ Token decryption failed: {e}, re-authenticating...")

```

connectors/toast/connector.py
```diff
@@ -612,10 +618,10 @@ def make_headers(conf, base_url, state, key):
...
 
     # Check if a valid token exists and is not expiring in the next hour
-    if (
-        "encrypted_token" in state
-        and "token_ttl" in state
-        and state["token_ttl"] > current_time + 3600
+    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
+        "encrypted_token" in state and
+        "token_ttl" in state and
+        state["token_ttl"] > current_time + 3600
```
fivetran-pranavtotala:
```suggestion
    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
```

connectors/toast/connector.py
```diff
@@ -612,10 +618,7 @@ def make_headers(conf, base_url, state, key):
     current_time = time.time()
 
     # Check if a valid token exists and is not expiring in the next hour
-    if (
-        "encrypted_token" in state
-        and "token_ttl" in state
-        and state["token_ttl"] > current_time + 3600
+    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
     ):
```
fivetran-pranavtotala:
```suggestion
```

connectors/toast/connector.py
```diff
@@ -612,18 +618,15 @@ def make_headers(conf, base_url, state, key):
...
-    ):
+    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
         try:
             auth_token = fernet.decrypt(state["encrypted_token"].encode()).decode()
             log.info("encrypted_token found with at least an hour left, reusing")
             return {"Authorization": f"Bearer {auth_token}", "Accept": "application/json"}, state
         except Exception as e:
             print(f"⚠️ Token decryption failed: {e}, re-authenticating...")
 
+
```
fivetran-pranavtotala:
```suggestion
```

connectors/toast/connector.py
```diff
@@ -612,18 +618,15 @@ def make_headers(conf, base_url, state, key):
     current_time = time.time()
 
     # Check if a valid token exists and is not expiring in the next hour
-    if (
-        "encrypted_token" in state
-        and "token_ttl" in state
-        and state["token_ttl"] > current_time + 3600
-    ):
+    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > current_time + 3600:
```
fivetran-pranavtotala:
```suggestion
    fut_time = current_time + 3600
    if "encrypted_token" in state and "token_ttl" in state and state["token_ttl"] > fut_time:
```

frederikkofoed:
@manjutapali is this good to be merged?

PR #540, opened 2026-03-17
feature(ehi): Adding high volume EHI example

fivetran-sahilkhirwal:
### Jira ticket
Closes https://fivetran.atlassian.net/browse/RD-1172756

### Description of Change
- Adding high volume EHI example

### Testing
Screenshot 2026-03-25 at 00 42 13
Screenshot 2026-03-25 at 00 52 25

Screenshot 2026-04-15 at 20 58 24

### Checklist

- [x] Tested the connector with `fivetran debug` command.
- [x] Added/Updated example-specific README.md file, see [the README template](https://github.com/fivetran/fivetran_connector_sdk/tree/main/template_connector/README_template.md) for the required structure and guidelines.
- [x] Followed Python Coding Standards, [refer here](https://github.com/fivetran/fivetran_connector_sdk/blob/main/PYTHON_CODING_STANDARDS.md)


fivetran-satvikpatil:
@codex review

connectors/ehi/high_volume_ehi/client.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        Acquire a connection from the pool as a context manager.
+        Always returns the connection to the pool in the finally block.
+        """
+        try:
+            conn = self._queue.get(timeout=timeout)
+        except queue.Empty:
+            raise RuntimeError(f"No connection available from pool within {timeout}s timeout")
+        try:
+            conn.ensure_open()
```
fivetran-satvikpatil, line 212: We don't handle case where this would throw exception, it won't yield anything.

connectors/ehi/high_volume_ehi/client.py
```diff
@@ -0,0 +1,209 @@
...
+        last_exc = None
+        for attempt in range(MAX_RETRIES):
+            try:
+                self.ensure_open()
+                cur = self._conn.cursor()
+                cur.execute(sql, params)
+                return cur
+            except Exception as exc:
+                if not _is_retryable_error(exc):
+                    log.severe(f"Non-retryable SQL error: {exc}")
```
fivetran-satvikpatil, line 148: Should we call `self._close()` and raise?

connectors/ehi/high_volume_ehi/models.py
```diff
@@ -0,0 +1,283 @@
...
+                if col.name.lower() == override.lower():
+                    log.fine(f"Using user-specified incremental column: {col.name}")
+                    return col
+            log.warning(
+                f"incremental_column '{override}' not found in column list; "
+                "falling back to auto-detection"
+            )
+
+        # Priority 2: known pattern match (case-insensitive)
+        col_map = {col.name.lower(): col for col in columns}
```
fivetran-satvikpatil, line 212: Do we really need to add this approach to check for the replication key? I think approach 1 should be enough.

fivetran-satvikpatil: Was there any requirement to add it?

connectors/ehi/high_volume_ehi/models.py
```diff
@@ -0,0 +1,283 @@
...
+    metadata. Also auto-detects a replication key for each table.
+    """
+
+    _METADATA_SQL = """
+        SELECT
+            c.COLUMN_NAME,
+            c.DATA_TYPE,
+            c.ORDINAL_POSITION,
+            COLUMNPROPERTY(
+                OBJECT_ID(c.TABLE_SCHEMA + '.' + c.TABLE_NAME),
```
fivetran-satvikpatil: This could cause problem if table_schema or table_name contains dot.

connectors/ehi/high_volume_ehi/models.py
```diff
@@ -0,0 +1,283 @@
...
+
+        return None
+
+    @staticmethod
+    def map_sql_type_to_python(sql_type: str):
+        """
+        Map a SQL Server DATA_TYPE string to a Python built-in type used by
+        convert_value() in readers.py.
+
+        Returns int, float, str, bool, bytes, or None.
```
fivetran-satvikpatil: We never return None

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,320 @@
...
+                sql = sql_next
+                params = (current_last, current_last, current_last_pk, self._batch_size)
+            else:
+                sql = sql_next
+                params = (current_last, self._batch_size)
+
+            with self._pool.acquire() as conn:
+                cur = conn.execute_with_retry(sql, params)
+                rows = cur.fetchall()
+                cur.close()
```
fivetran-satvikpatil: Add this in finally block, it won't run if cur.fetchall() fails

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,586 @@
...
+            return prior_mode  # resume interrupted full load
+        return "incremental" if has_replication_key else _full_mode()
+
+    if prior_mode == "incremental":
+        return "incremental"
+
+    return _full_mode()
+
+
+def _save_checkpoint(
```
fivetran-satvikpatil, line 190: Should we use any thread lock to ensure we don't face any race condition during checkpoint?

connectors/ehi/high_volume_ehi/models.py
fivetran-sahilkhirwal: The goal is to make this example robust where it can handle multiple scenarios without requiring code changes . It is just to give the user more options

connectors/ehi/high_volume_ehi/connector.py
fivetran-sahilkhirwal: Each thread is handling its own table. It will not cause any race condition

connectors/ehi/high_volume_ehi/client.py
```diff
@@ -0,0 +1,213 @@
...
+        except queue.Empty:
+            raise RuntimeError(f"No connection available from pool within {timeout}s timeout")
+        try:
+            conn.ensure_open()
+            yield conn
+        except Exception as exc:
+            log.warning(f"Connection error during acquire: {exc}")
+            raise
+        finally:
+            self._queue.put(conn)
```
fivetran-satvikpatil, line 219: The finally block is called even in case of failure. Shouldn't we just add it to queue when conn is opened correctly?

connectors/ehi/high_volume_ehi/client.py
```diff
@@ -0,0 +1,213 @@
...
+        self._queue = queue.Queue(maxsize=size)
+        self._all: list = []
+        try:
+            for _ in range(size):
+                conn = MSSQLConnection(configuration)
+                conn.ensure_open()
+                self._all.append(conn)
+                self._queue.put(conn)
+        except Exception:
+            self.close_all()
```
fivetran-satvikpatil: Even if one of the connection fails, we are closing all the connections. We could move this try-catch inside the for loop.

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,588 @@
...
+    table_include, table_exclude = _parse_table_list()
+    pool_size = MAX_WORKERS + 2
+    pool = ConnectionPool(configuration=configuration, size=pool_size)
+
+    try:
+        # Reuse schema cache when schema() was called earlier in this process
+        if _CACHED_TABLE_SCHEMAS:
+            table_schemas = _CACHED_TABLE_SCHEMAS
+        else:
+            detector = SchemaDetector(pool)
```
fivetran-satvikpatil: We could extract this else part logic in separate method and reuse it in schema method as well.

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,588 @@
...
+        table_state: dict with existing state for this table, to be updated and saved
+        pool: ConnectionPool to acquire SQL connections from
+        batch_size: number of rows to fetch per batch from the database
+        checkpoint_interval: number of rows to process between checkpoints
+    """
+    table_name = table_schema.table_name
+    has_repl_key = table_schema.replication_key is not None
+    mode = "full" if has_repl_key else "full_offset"
+
+    last_marker = (
```
fivetran-satvikpatil: This value could be None, but if there is no data, we are trying to checkpoint the state with None value itself. Please handle this case.

connectors/ehi/high_volume_ehi/connector.py
fivetran-satvikpatil: It is true, they are not updating the similar key, but consider what we do when a checkpoint is called, we try getting all the records from the queue and create a gRPC message to send it to the client. 
Now, what if we are creating this gRPC message, but during that time, one of the keys in the state was getting updated, it could have an incomplete value in the gRPC message, which would cause a problem.

connectors/ehi/high_volume_ehi/constants.py
```diff
@@ -0,0 +1,57 @@
...
+# extra check on the native error number embedded in the message string.
+LOCK_TIMEOUT_NATIVE_ERROR = "1222"
+
+DEFAULT_TABLE_LIST = ""  # Empty = sync every table in the schema
+DEFAULT_TABLE_EXCLUSION_LIST = ""  # Empty = exclude nothing
+DEFAULT_USE_PK_TIEBREAK = False  # Set True for tables with duplicate replication key values
+DEFAULT_FORCE_FULL_RESYNC = False  # Set True to discard state and re-sync from scratch
+
+KNOWN_REPLICATION_KEY_PATTERNS = [
+    "_LastUpdatedInstant",
```
fivetran-satvikpatil, line 44: Do we also need to add: `last_updated_instant`?

connectors/ehi/high_volume_ehi/models.py
```diff
@@ -0,0 +1,287 @@
...
+        # Priority 2: known pattern match (case-insensitive)
+        col_map = {col.name.lower(): col for col in columns}
+        for pattern in KNOWN_REPLICATION_KEY_PATTERNS:
+            if pattern.lower() in col_map:
+                matched_col = col_map[pattern.lower()]
+                log.fine(f"Replication key matched pattern '{pattern}': {matched_col.name}")
+                return matched_col
+
+        # Priority 3: first temporal column
+        datetime_types = {"datetime2", "datetime", "datetimeoffset", "smalldatetime", "date"}
```
fivetran-satvikpatil: Should this be in constants.py?

connectors/ehi/high_volume_ehi/models.py
```diff
@@ -0,0 +1,287 @@
...
+        return [col.name for col in self.columns if col.is_primary_key]
+
+    @property
+    def selectable_columns(self) -> list:
+        """
+        Return columns safe to SELECT: not computed and with a known python_type.
+        Computed columns are excluded because SQL Server rejects them in explicit
+        column lists under certain contexts.
+        """
+        return [col for col in self.columns if not col.is_computed and col.python_type is not None]
```
fivetran-satvikpatil, line 66: Do we handle the None case for is_computed? How should this work in that case?

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,326 @@
...
+            )
+        else:
+            sql_first = (
+                f"SELECT {col_sql} FROM {schema_table} WITH (NOLOCK) "
+                f"WHERE [{repl_col}] IS NOT NULL "
+                f"ORDER BY [{repl_col}] {fetch}"
+            )
+            sql_next = (
+                f"SELECT {col_sql} FROM {schema_table} WITH (NOLOCK) "
+                f"WHERE [{repl_col}] > ? "
```
fivetran-satvikpatil, line 229: Don't we need `is not null` check here?

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,326 @@
...
+                record = {
+                    col.name: convert_value(raw, col.python_type)
+                    for col, raw in zip(sel_cols, row)
+                }
+                batch.append(record)
+                rk_raw = record.get(repl_col)
+                if rk_raw is not None:
+                    last_value = rk_raw
+                if tiebreak_pk:
+                    pk_raw = record.get(tiebreak_pk)
```
fivetran-satvikpatil: This could cause an infinite loop. What if repl_col is None for all and even tiebreak_pk is None for all. In that case, it would always run line 150

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,326 @@
...
+            last_value = current_last
+            batch = []
+            for row in rows:
+                record = {
+                    col.name: convert_value(raw, col.python_type)
+                    for col, raw in zip(sel_cols, row)
+                }
+                batch.append(record)
+                rk_raw = record.get(repl_col)
+                if rk_raw is not None:
```
fivetran-satvikpatil: Same here.

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,326 @@
...
+    if python_type is float:
+        return float(value)
+    if python_type is bool:
+        return bool(value)
+    if python_type is bytes:
+        # Binary, spatial, rowversion — encode as base64 ASCII string
+        if isinstance(value, (bytes, bytearray, memoryview)):
+            return base64.b64encode(bytes(value)).decode("ascii")
+        return None
+    return None  # unknown / None python_type — skip column
```
fivetran-satvikpatil: Should we return str(value) by default?

connectors/ehi/high_volume_ehi/constants.py
fivetran-sahilkhirwal: This is a basic set of replication key that might be present. The users can easily add any other pattern as required

connectors/ehi/high_volume_ehi/models.py
fivetran-sahilkhirwal: The value is stored as bool(is_computed)  
  at line 135 of models.py:                                                     
  is_computed=bool(is_computed),                                         
  In Python, bool(None) = False, so a NULL from SQL Server is treated as "not   
  computed". This is conservative (it includes the column in the SELECT list,
  which may fail if it is actually computed but returns NULL for the property).

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,610 @@
...
+import traceback
+
+# For concurrent execution of table syncs in multiple threads
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+# For timestamping sync start and completion times in ISO 8601 format
+from datetime import datetime, timezone
+
+# Import required classes from fivetran_connector_sdk
+# For supporting Connector operations like Update() and Schema()
```
fivetran-JenasVimal:
```suggestion
```

connectors/ehi/high_volume_ehi/constants.py
```diff
@@ -0,0 +1,62 @@
+"""
+Constants for EHI connector.
+Modify these values to tune performance and behaviour for your environment.
+"""
+
```
fivetran-JenasVimal, line 5: please ensure it is the meeting the python coding standards 
`__MAX_RETRIES = 5  # Maximum retry attempts on transient SQL errors

__BASE_RETRY_DELAY = 5.0  # Initial backoff delay in seconds (doubles each attempt)

__MAX_RETRY_DELAY = 300.0  # Ceiling for exponential backoff (5 minutes)

__CONNECTION_TIMEOUT_HOURS = 8  # Hours before proactively cycling a connection`

connectors/ehi/high_volume_ehi/readers.py
```diff
@@ -0,0 +1,337 @@
...
+        # Pre-build SQL template.
+        # WHERE [repl_col] > ? implicitly excludes NULL replication-key rows —
+        # SQL NULL comparisons always evaluate to UNKNOWN/FALSE, so no explicit IS NOT NULL needed.
+        sql = (
+            f"SELECT {col_sql} FROM {schema_table} WITH (NOLOCK) "
+            f"WHERE [{repl_col}] > ? "
+            f"ORDER BY [{repl_col}] "
+            f"OFFSET 0 ROWS FETCH NEXT ? ROWS ONLY "
+            f"OPTION (FAST {self._batch_size})"
+        )
```
fivetran-JenasVimal: Suppose 2,000 rows are inserted with identical `repl_col = 2024-01-01 00:00:00`. Batch 1 returns 1,000 of them `current_last becomes` 2024-01-01. 

Batch 2 runs WHERE repl_col > 2024-01-01 — the other 1,000 rows at that exact timestamp are
permanently skipped. No warning is logged.

Ig can you some kind of composite PK to avoid this

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,610 @@
...
+    if DEFAULT_TABLE_LIST:
+        included_names = [
+            name.strip()
+            for name in DEFAULT_TABLE_LIST.split(",")
+            if name.strip() and name.strip().lower() not in excluded_names
+        ]
+        return (included_names or None), excluded_names
+
+    # No explicit include list — discover all tables; exclusion applied later
+    return None, excluded_names
```
fivetran-JenasVimal, line 109: If you explicitly set `DEFAULT_TABLE_LIST = "orders"` and `DEFAULT_TABLE_EXCLUSION_LIST = "orders"` — all explicit tables are excluded — included_names becomes []. 
Then [] or None = None, which the rest of the code interprets as "sync every table in the schema". The intended result was zero tables. The connector silently syncs everything instead of nothing.


Adding code flow 
update() / schema()
        ↓
_parse_table_list()
        ↓
(included_names, excluded_names)
        ↓
_discover_table_schemas()
        ↓
SchemaDetector.detect_all_tables() // Here is syncs everything

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,610 @@
...
+    https://fivetran.com/docs/connectors/connector-sdk/technical-reference#update
+    Args:
+        configuration: A dictionary containing connection details
+        state: A dictionary containing state information from previous runs
+        The state dictionary is empty for the first sync or for any full re-sync
+    """
+    log.warning("Example: Connectors - EHI")
+    schema_name = configuration.get("mssql_schema", "dbo")
+    table_include, table_exclude = _parse_table_list()
+    pool_size = MAX_WORKERS + 2
```
fivetran-JenasVimal, line 605: please add comments as to why we add `+2`

fivetran-JenasVimal (changes requested):
Please address the comments

connectors/ehi/high_volume_ehi/connector.py
fivetran-sahilkhirwal: This is a very rare case where the table list and exclude list is same. In such cases the connector be just paused because no data is required

connectors/ehi/high_volume_ehi/constants.py
fivetran-sahilkhirwal: These are in a separate constant file. This change is not required

connectors/ehi/high_volume_ehi/constants.py
```diff
@@ -0,0 +1,62 @@
...
+)
+
+# SQL Server native error 1222 = lock request time out period exceeded.
+# It arrives with SQLSTATE HY000 (general database error), so we need an
+# extra check on the native error number embedded in the message string.
+LOCK_TIMEOUT_NATIVE_ERROR = "1222"
+
+DEFAULT_TABLE_LIST = ""  # Empty = sync every table in the schema
+DEFAULT_TABLE_EXCLUSION_LIST = ""  # Empty = exclude nothing
+DEFAULT_USE_PK_TIEBREAK = False  # Set True for tables with duplicate replication key values
```
fivetran-JenasVimal: Please add a conditional statement for this

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+    missing_fields = [
+        field_name
+        for field_name in required_fields
+        if not configuration.get(field_name, "").strip()
+    ]
+    if missing_fields:
+        raise ValueError(
+            f"Missing required configuration fields: {missing_fields}. "
+            "Check configuration.json."
+        )
```
fivetran-JenasVimal, line 77: Validate the field type as well , to prevent wrong imports

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+        FROM sys.indexes    i
+        JOIN sys.partitions p
+            ON  i.object_id = p.object_id
+            AND i.index_id  = p.index_id
+        WHERE i.index_id <= 1
+          AND OBJECT_SCHEMA_NAME(i.object_id) = ?
+        GROUP BY i.object_id
+    """
+    with pool.acquire() as conn:
+        cursor = conn.execute_with_retry(sql, (schema_name,))
```
fivetran-JenasVimal: Retrieves all the tables in that schema , and not the respective table name , might cause problems later

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+    size_map = {row[0]: int(row[1] or 0) for row in rows}
+    return {name: size_map.get(name, 0) for name in table_names}
+
+
+def _determine_mode(table_state: dict, table_schema: TableSchema) -> str:
+    """
+    Decide whether to run a full load or incremental load for a table.
+    Decision tree:
+      DEFAULT_FORCE_FULL_RESYNC=True → 'full' or 'full_offset'
+      No prior state                 → 'full' or 'full_offset'
```
fivetran-JenasVimal: idt full is necessary , do check !

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+            # The first argument is the name of the destination table.
+            # The second argument is a dictionary containing the record to be upserted.
+            op.upsert(table_name, row)
+
+        batch_len = len(batch)
+        rows_synced += batch_len
+        rows_since_checkpoint += batch_len
+        last_marker = progress_marker
+
+        if rows_since_checkpoint >= checkpoint_interval:
```
fivetran-JenasVimal: can we not checkpoint every '%' of the checkpoint_interval

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+        f"{table_name}: starting {mode} load "
+        f"(rows_so_far={rows_synced}, last_marker={last_marker})"
+    )
+
+    for batch, progress_marker in reader.read_batches():
+        for row in batch:
+            # The 'upsert' operation is used to insert or update data in the destination table.
+            # The first argument is the name of the destination table.
+            # The second argument is a dictionary containing the record to be upserted.
+            op.upsert(table_name, row)
```
fivetran-JenasVimal: if batch is 10,000 , then 10,000 upserts seem kinda inefficient , i think we can put them in record , because even if it fails it still wouldnt have been checkpointed , would be easier to see in logs ig

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+    current_last = last_replication_value
+
+    log.info(f"{table_name}: starting incremental sync from {last_replication_value}")
+
+    for batch, last_seen in reader.read_upsert_batches():
+        for row in batch:
+            # The 'upsert' operation is used to insert or update data in the destination table.
+            # The first argument is the name of the destination table.
+            # The second argument is a dictionary containing the record to be upserted.
+            op.upsert(table_name, row)
```
fivetran-JenasVimal, line 432: same as above

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+
+    pool = ConnectionPool(configuration=configuration, size=1)
+    try:
+        table_schemas = _discover_table_schemas(
+            pool, schema_name, table_include, table_exclude, config=configuration, max_workers=1
+        )
+    finally:
+        pool.close_all()
+
+    _CACHED_TABLE_SCHEMAS = table_schemas  # cache for update() reuse
```
fivetran-JenasVimal, line 578: whats the purpose ?

fivetran-sahilkhirwal: So that we do not build schema twice -> Once in schema() and again in update()
Saves compute and time

connectors/ehi/high_volume_ehi/readers.py
fivetran-satvikpatil: @fivetran-sahilkhirwal ^^

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,616 @@
...
+        else frozenset()
+    )
+
+    if DEFAULT_TABLE_LIST:
+        included_names = [
+            name.strip()
+            for name in DEFAULT_TABLE_LIST.split(",")
+            if name.strip() and name.strip().lower() not in excluded_names
+        ]
+        return (included_names or None), excluded_names
```
fivetran-satvikpatil: This returns None for tables_included and then we pass this to `_discover_table_schemas` which passes this to `detect_all_tables`, now here if the table_names is None, we are actually listing all the tables. So isn't it an incorrect behaviour if we don't want to include any table but just for passing None, we are fetching all tables.

fivetran-satvikpatil (dismissed):
Please address one comment, else LGTM. Also please add testing details again after the changes.

connectors/ehi/high_volume_ehi/readers.py
fivetran-sahilkhirwal: SQL uses three-valued logic (TRUE / FALSE / UNKNOWN). Any comparison against  
  NULL — including >, <, =, != — evaluates to UNKNOWN, not TRUE or FALSE. The
  WHERE clause only passes rows where the predicate is TRUE, so UNKNOWN is      
  treated as FALSE and those rows are automatically excluded.

fivetran-satvikpatil (dismissed):

fivetran-JenasVimal:
fix DATETIME2(7) infinite loop in KeysetReader due to precision loss                                                                                                                           
                                                                                                                                                                                             
  pyodbc truncates DATETIME2(7) to Python datetime microseconds (6 digits),                                                                                                                      
  making the cursor always slightly less than the actual row value. The same                                                                                                                   
  row satisfies > cursor on every page, looping forever.                                                                                                                                         
                                                                                                                                                                                                 
  Fix: add CONVERT(NVARCHAR(50), [repl_col], 127) as an extra SELECT column                                                                                                                      
  for datetime2/datetimeoffset replication keys. SQL Server style 127 returns                                                                                                                    
  full ISO 8601 with all 7 digits. Use this string directly as the cursor;                                                                                                                       
  zip(sel_cols, row) stops at len(sel_cols) so the extra column is never                                                                                                                         
  upserted.                                          
  
  
###   Historical sync 
  
<img width="1456" height="378" alt="image" src="https://github.com/user-attachments/assets/c8d2199b-22f7-49c4-a46c-435eba1fc64a" />


### Incremental sync 
<img width="1515" height="357" alt="image" src="https://github.com/user-attachments/assets/cfa34e79-d7f4-4723-b0dc-b199ca280b2d" />

fivetran-satvikpatil:
@codex review

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,687 @@
...
+                table_name = future_to_table[future]
+                try:
+                    future.result()
+                    log.info(f"{table_name}: thread finished successfully")
+                except Exception as exc:
+                    log.severe(f"{table_name}: thread raised unhandled exception: {exc}")
+                    failed_tables.append(table_name)
+
+        if failed_tables:
+            log.warning(f"The following tables failed to sync: {failed_tables}")
```
fivetran-satvikpatil, line 654: make this severe and also, if all tables failed, we can just throw RuntimeError

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,687 @@
...
+            f"{table_name}: starting full keyset load "
+            f"(rows_so_far={rows_synced}, last_marker={last_marker})"
+        )
+        for batch, progress_marker, progress_pk_marker in reader.read_batches():
+            for row in batch:
+                op.upsert(table_name, row)
+            rows_synced += len(batch)
+            last_marker = progress_marker
+            last_pk_marker = progress_pk_marker
+            if rows_synced % checkpoint_interval == 0:
```
fivetran-satvikpatil, line 287: We are not actually incrementing the rows_synced based on each row upsertion, but we are increasing by len(batch). Now this could just cross the 25k record, and it won't checkpoint because it is not exactly divisible by 25k. 
Example: 
```
rows_synced starts at 25,500  (loaded from state, line 263)                                                                                                                                                    
                                                                                                                                                                                                                 
  After batch 1:  26,500  →  26500 % 25000 = 1500  ✗                                                                                                                                                             
  After batch 2:  27,500  →  27500 % 25000 = 2500  ✗                                                                                                                                                             
  ...                                                                                                                                                                                                            
  After batch 24: 49,500  →  49500 % 25000 = 24500 ✗                                                                                                                                                           
  After batch 25: 50,500  →  50500 % 25000 =  500  ✗  ← jumped OVER 50,000                                                                                                                                       
  After batch 26: 51,500  →  51500 % 25000 = 1500  ✗
```

connectors/ehi/high_volume_ehi/connector.py
```diff
@@ -0,0 +1,687 @@
...
+
+    if not table_state:
+        return "full"
+
+    prior_mode = table_state.get("mode", "full")
+
+    if prior_mode == "full":
+        if table_state.get("sync_completed_at") is None:
+            return "full"
+        return "incremental" if has_replication_key else "full"
```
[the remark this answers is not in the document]
fivetran-JenasVimal: good catch. ,can resolve this by add another condition that checks if the replication key is null or not

